PR #5, opened 2020-12-04
add LongTensor

vieting:
Add `LongTensor`. Not sure if we need more than that.

### DIFF
--- a/pytorch_to_returnn/torch/__init__.py
+++ b/pytorch_to_returnn/torch/__init__.py
@@ -1,5 +1,5 @@
 
-from .tensor import Tensor
+from .tensor import *
 from ._C import *
 from .nn.functional import *
 from .serialization import load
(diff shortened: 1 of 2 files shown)


pytorch_to_returnn/torch/tensor.py
```diff
@@ -145,3 +145,8 @@ def __mul__(self, other):
   def __truediv__(self, other):
     from .nn.functional import truediv
     return truediv(self, other)
+
+
+class LongTensor(Tensor):
+  def __init__(self, **kwargs):
+    super(LongTensor, self).__init__(dtype="float64", **kwargs)
```
albertz, line 152: Wrong dtype. Should be int64, right?
```suggestion
    super(LongTensor, self).__init__(dtype="int64", **kwargs)
```

pytorch_to_returnn/torch/__init__.py
```diff
@@ -1,5 +1,5 @@
 
-from .tensor import Tensor
+from .tensor import Tensor, LongTensor
```
albertz, line 2:
```suggestion
from .tensor import *
```
Better?

pytorch_to_returnn/torch/tensor.py
vieting: Oh yes, that's right of course.

albertz (approved):

PR #8, opened 2020-12-04
fix of default parameters in ConvTranspose2d

vieting:
Default parameter shapes were still 1D

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/conv.py
+++ b/pytorch_to_returnn/torch/nn/modules/conv.py
@@ -202,11 +202,6 @@ class ConvTranspose1d(_ConvTransposeNd):
       dilation: _size_1_t = 1,
       padding_mode: str = 'zeros'
   ):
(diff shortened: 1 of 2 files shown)


albertz:
We have:
```
_size_2_t = _scalar_or_tuple_2_t[int]
```
So according to the doc, this should work. Why doesn't it?

albertz:
Also, I think it works like this in Torch as well. I.e. you can write `ConvTransposed2d(2, 3, 4, 5)`.
So this must work, otherwise there is some bug.

vieting:
Yes, you're right. I think the problem was not in the default values but in parsing them as tuples with single entry. I removed this and added a test case. Also, I think we can remove the `_single` statements for the 1d transposed convolution. This is basically a duplicate of what is done in the parent class `_ConvNd` with `_ntuple` anyway.

albertz (approved):

PR #10, opened 2020-12-07
add tensor.expand

vieting:
Add `tensor.expand`. To do this, first reshape to get additional axes if necessary and then use RETURNN's `TileLayer` to expand to larger size.

### DIFF
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -186,6 +186,34 @@ def tensorflow_transpose(input: Tensor, perm: Optional[Union[Dict[int, int], Tup
   return modules.Transpose(perm=perm)(input)
 
 
+def expand(input: Tensor, *sizes: _size):
+  if sizes and isinstance(sizes[0], (tuple, list)):
+    assert len(sizes) == 1
+    shape = tuple(sizes[0])
+  else:
+    shape = sizes
+  assert isinstance(shape, tuple)
+
+  # reshape to add new dims
+  if len(shape) != len(input.shape):
+    assert len(shape) >= len(input.shape), "Cannot reduce dimensions"
+    assert -1 not in [d for d in shape[:-len(input.shape)]], "Size -1 is not allowed for expanded dims"
+    exp_shape = (1,) * (len(shape) - len(input.shape)) + input.shape[-len(input.shape):]
+    input = reshape(input, shape=exp_shape)
+
+  # tile to expand to larger size
+  multiples = {}
+  for axis, multiple in enumerate(shape):
+    if input.shape[axis] > 1:
+      assert multiple == -1 or multiple == input.shape[axis], \
+        f"Expanded size ({multiple}) must match existing size ({input.shape[axis]}) at non-singleton dimension ({axis})"
+      multiple = 1
+    if multiple == -1:
+      multiple = 1  # -1 as the size for a dim means not changing the size of that dim
+    multiples[axis] = multiple
+  return modules.Tile(multiples=multiples).as_returnn_torch_functional()(input)
+
+
 def pad(input: Tensor, pad, mode='constant', value=0) -> Tensor:
   return modules.GenericPadNd(padding=pad, mode=mode, value=value).as_returnn_torch_functional()(input)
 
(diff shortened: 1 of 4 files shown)


pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -186,6 +186,34 @@ def tensorflow_transpose(input: Tensor, perm: Optional[Union[Dict[int, int], Tup
...
+    assert len(sizes) == 1
+    shape = tuple(sizes[0])
+  else:
+    shape = sizes
+  assert isinstance(shape, tuple)
+
+  # reshape to add new dims
+  if len(shape) != len(input.shape):
+    assert len(shape) >= len(input.shape), "Cannot reduce dimensions"
+    assert not -1 in [d for d in shape[:-len(input.shape)]], "Size -1 is not allowed for expanded dims"
```
albertz, line 200:
```suggestion
    assert -1 not in [d for d in shape[:-len(input.shape)]], "Size -1 is not allowed for expanded dims"
```
This is canonical Python code (PEP8).
(PyCharm should also tell you that, right?)

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -186,6 +186,34 @@ def tensorflow_transpose(input: Tensor, perm: Optional[Union[Dict[int, int], Tup
...
+    exp_shape = (1,) * (len(shape) - len(input.shape)) + input.shape[-len(input.shape):]
+    input = reshape(input, shape=exp_shape)
+
+  # tile to expand to larger size
+  multiples = dict()
```
albertz, line 205:
```suggestion
  multiples = {}
```

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -212,6 +212,23 @@ def create_returnn_layer_dict(self, *inputs: Tensor) -> Dict[str, Any]:
...
+
+  def __init__(self, multiples: Dict[str, int]):
+    super(Tile, self).__init__()
+    assert isinstance(multiples, dict)
+    self.multiples = multiples
+
+  def create_returnn_layer_dict(self, *inputs: Tensor) -> Dict[str, Any]:
+    return {
+      "class": "tile", "multiples": self.multiples,
+      "from": [self._get_input_layer_name(x) for x in inputs]}
```
albertz, line 229: Why do you support having multiple inputs here?
A module should be simple, and not do extra logic.
```suggestion
  def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
    return {
      "class": "tile", "multiples": self.multiples,
      "from": self._get_input_layer_name(input)}
```

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -212,6 +212,23 @@ def create_returnn_layer_dict(self, *inputs: Tensor) -> Dict[str, Any]:
       "from": [self._get_input_layer_name(x) for x in inputs]}
 
 
+class Tile(Module):
+  """
+  Wraps RETURNN TileLayer.
+  """
+  is_original_torch_module = False
+
+  def __init__(self, multiples: Dict[str, int]):
```
albertz, line 221: A PyTorch module should not mix with RETURNN logic here. The keys are RETURNN axes. That doesn't make sense.
The call to `get_returnn_axis_description` should actually be in `create_returnn_layer_dict` below.

albertz (changes requested):
See comments.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -212,6 +212,26 @@ def create_returnn_layer_dict(self, *inputs: Tensor) -> Dict[str, Any]:
...
+  is_original_torch_module = False
+
+  def __init__(self, multiples: Dict[int, int]):
+    super(Tile, self).__init__()
+    assert isinstance(multiples, dict)
+    self.multiples = multiples
+
+  def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
+    multiples = {
+      input.returnn_naming_entry.get_returnn_axis_description(torch_axis=axis): multiple
```
albertz, line 228:
```suggestion
      self._get_input_axis_to_returnn(input, axis=axis): multiple
```

albertz (approved):

PR #11, opened 2020-12-08
Add more tensor ops

vieting:
Add more tensor ops, namely `tensor.flatten()` and `tensor.resize()` (currently really minimal).


vieting:
Also added `tensor.int()`

PR #12, opened 2020-12-08
add tensor greater equal op

vieting:
Add tensor greater equal op. Other comparisons can be added analogously.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -48,6 +48,22 @@ class BinaryOperator(Module):
       "from": [self._get_input_layer_name(input) for input in inputs]}
 
 
+class ComparisonOperator(BinaryOperator):
+  is_original_torch_module = False
+
+  def __init__(self, kind: str):
+    """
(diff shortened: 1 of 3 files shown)


pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -48,6 +48,22 @@ def create_returnn_layer_dict(self, *inputs: Tensor):
       "from": [self._get_input_layer_name(input) for input in inputs]}
 
 
+class ComparisonOperator(BinaryOperator):
+  is_original_torch_module = False
+
+  def __init__(self, kind: str):
+    """
+    :param str kind: e.g. “equal”, “greater”, “less” or other supported TF comparison ops
```
albertz, line 56:
```suggestion
    :param str kind: e.g. "equal", "greater", "less" or other supported TF comparison ops
```

JackTemaki (approved):
See Alberts comment, otherwise good.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: Ups, how did that happen? :D

PR #16, opened 2020-12-11
Reduced printing of numpy arrays

vieting:
It's maybe personal preference, but I don't like if all values for all numpy arrays are always printed, especially when a lot of `ConstantLayer`s are used. I therefore added a threshold and if a numpy array contains more values, only an abbreviated string is printed.

### DIFF
--- a/pytorch_to_returnn/naming/call.py
+++ b/pytorch_to_returnn/naming/call.py
@@ -9,6 +9,7 @@ from . import module as _module
 from . import naming as _naming
 from . import tensor as _tensor
 from . import namespace as _namespace
+from ..pprint import pprint
 
 
 class CallEntry:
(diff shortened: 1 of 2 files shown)


albertz:
I'm not sure on this. This `pprint` is not intended for debugging output (although also useful for that), but for printing the RETURNN net dict such that you can copy & paste it into your RETURNN config and it will directly work.

You are right that this deviates from the original `__repr__` or `__str__` of a Numpy array (and the original `pprint` also just uses the original `__repr__` and has no extra logic for Numpy arrays).

But there is already the second problem with your approach, that this is inconsistent as well to the original `ndarray.__repr__`. Numpy has options for this, specifically e.g. via [`numpy.set_printoptions`](numpy.set_printoptions).

I wonder also, you changed a `print` to `pprint` here. But the original `print` already used the original Numpy mechanism, which should also have done that?

These `print`s (all via the converter) should anyway be cleaned up. We might use the [Python `logging` mechanism](https://docs.python.org/3/howto/logging.html). This was/is mostly for debugging. You should be able to control that with a verbosity setting.

vieting:
I tried `numpy.set_printoptions` but still found that quite verbose. But yes, for my case now the easiest thing might be to reduce the verbosity level.

PR #18, opened 2020-12-15
add group norm

vieting:
Add PyTorch's group norm using `NormLayer` in RETURNN.

### DIFF
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -344,3 +344,16 @@ def norm(input: Tensor,
 
 def norm_except_dim(v: Tensor, pow: int = 2, dim: int = 0) -> Tensor:
   return modules.Norm(p=pow, axes=[i for i in range(v.dim()) if i != dim], keepdims=True)(v)
+
+
+def group_norm(input, num_groups, weight=None, bias=None, eps=1e-5) -> Tensor:
+  if weight is None:
+    assert bias is None
+    module = modules.GroupNorm(num_groups=num_groups, num_channels=0, eps=eps, affine=False)
+  else:
+    assert bias is not None
+    assert weight.shape == bias.shape
+    module = modules.GroupNorm(num_groups=num_groups, num_channels=weight.shape[0], eps=eps)
+    module.register_parameter("weight", weight)
+    module.register_parameter("bias", bias)
+  return module.as_returnn_torch_functional()(input)
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -346,14 +346,14 @@ def norm_except_dim(v: Tensor, pow: int = 2, dim: int = 0) -> Tensor:
   return modules.Norm(p=pow, axes=[i for i in range(v.dim()) if i != dim], keepdims=True)(v)
 
 
-def group_norm(input, num_groups, weight=None, bias=None, eps=1e-5) -> Tensor:
-  if weight is None:
-    assert bias is None
-    module = modules.GroupNorm(num_groups=num_groups, num_channels=0, eps=eps, affine=False)
-  else:
-    assert bias is not None
-    assert weight.shape == bias.shape
-    module = modules.GroupNorm(num_groups=num_groups, num_channels=weight.shape[0], eps=eps)
-    module.register_parameter("weight", weight)
-    module.register_parameter("bias", bias)
-  return module.as_returnn_torch_functional()(input)
+def group_norm(input: Tensor, num_groups: int, weight: Optional[Tensor] = None, bias: Optional[Tensor] = None,
+               eps: float = 1e-5) -> Tensor:
+  module = modules.GroupNorm(num_groups=num_groups, num_channels=0, eps=eps, affine=False)
+  out = module.as_returnn_torch_functional()(input)
+  assert out.shape[1] == weight.shape[0], "data should be of shape (B, F, *) and weights should be of shape (F,)"
+  assert out.shape[1] == bias.shape[0], "data should be of shape (B, F, *) and bias should be of shape (F,)"
+  if weight is not None:
+    out *= weight.view(1, -1, 1)
+  if bias is not None:
+    out += bias.view(1, -1, 1)
+  return out
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -350,10 +350,10 @@ def group_norm(input: Tensor, num_groups: int, weight: Optional[Tensor] = None, 
                eps: float = 1e-5) -> Tensor:
   module = modules.GroupNorm(num_groups=num_groups, num_channels=0, eps=eps, affine=False)
   out = module.as_returnn_torch_functional()(input)
-  assert out.shape[1] == weight.shape[0], "data should be of shape (B, F, *) and weights should be of shape (F,)"
-  assert out.shape[1] == bias.shape[0], "data should be of shape (B, F, *) and bias should be of shape (F,)"
   if weight is not None:
+    assert (out.shape[1],) == weight.shape, "data should be of shape (B, F, *) and weights should be of shape (F,)"
     out *= weight.view(1, -1, 1)
   if bias is not None:
+    assert (out.shape[1],) == bias.shape, "data should be of shape (B, F, *) and bias should be of shape (F,)"
     out += bias.view(1, -1, 1)
   return out
--- a/pytorch_to_returnn/torch/nn/modules/normalization.py
+++ b/pytorch_to_returnn/torch/nn/modules/normalization.py
@@ -4,7 +4,7 @@ import math
 import numbers
 from typing import Dict, Any, Union, List
 import tensorflow as tf
-from returnn.tf.layers.basic import LayerNormLayer
+from returnn.tf.layers.basic import LayerNormLayer, NormLayer, SubnetworkLayer
 from .module import Module
 from ..parameter import Parameter
 from ...tensor import Tensor, Size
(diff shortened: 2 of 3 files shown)


tests/test_layers.py
```diff
@@ -314,6 +314,26 @@ def model_func(wrapped_import, inputs: torch.Tensor):
   numpy.testing.assert_allclose(mean_returnn, mean_torch, rtol=0, atol=1e-5)
 
 
+def test_group_norm():
+  n_batch, n_time = 4, 20
+  for num_groups, num_channels in [(1, 512), (512, 512)]:
```
albertz, line 319: You should use smaller numbers for a test (like 5 or so).

pytorch_to_returnn/torch/nn/modules/normalization.py
```diff
@@ -53,6 +53,51 @@ def import_params_torch_to_returnn(self, *, layer: LayerNormLayer, torch_module:
...
+  def reset_parameters(self) -> None:
+      if self.affine:
+          init.ones_(self.weight)
+          init.zeros_(self.bias)
+
+  def extra_repr(self) -> str:
+      return '{num_groups}, {num_channels}, eps={eps}, affine={affine}'.format(**self.__dict__)
+
+  def import_params_torch_to_returnn(self, *, layer: NormLayer, torch_module: GroupNorm):
+    if isinstance(layer, SubnetworkLayer):
```
albertz, line 81: I don't exactly understand, how can this happen?
This should never happen. The layer it gets here should be exactly the layer created via `create_returnn_layer_dict`. If this is not the case, sth is wrong.

pytorch_to_returnn/torch/nn/modules/normalization.py
```diff
@@ -53,6 +53,51 @@ def import_params_torch_to_returnn(self, *, layer: LayerNormLayer, torch_module:
...
+        self.bias = Parameter(Tensor(num_channels))
+    else:
+        self.register_parameter('weight', None)
+        self.register_parameter('bias', None)
+    self.reset_parameters()
+
+  def reset_parameters(self) -> None:
+      if self.affine:
+          init.ones_(self.weight)
+          init.zeros_(self.bias)
```
albertz, line 75: The indent is wrong (4 spaces but should be 2).

pytorch_to_returnn/torch/nn/modules/normalization.py
```diff
@@ -53,6 +53,51 @@ def import_params_torch_to_returnn(self, *, layer: LayerNormLayer, torch_module:
...
+    self.num_groups = num_groups
+    self.num_channels = num_channels
+    self.eps = eps
+    self.affine = affine
+    if self.affine:
+        self.weight = Parameter(Tensor(num_channels))
+        self.bias = Parameter(Tensor(num_channels))
+    else:
+        self.register_parameter('weight', None)
+        self.register_parameter('bias', None)
```
albertz, line 69: Indent wrong.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -344,3 +344,16 @@ def norm(input: Tensor,
...
+def group_norm(input, num_groups, weight=None, bias=None, eps=1e-5) -> Tensor:
+  if weight is None:
+    assert bias is None
+    module = modules.GroupNorm(num_groups=num_groups, num_channels=0, eps=eps, affine=False)
+  else:
+    assert bias is not None
+    assert weight.shape == bias.shape
+    module = modules.GroupNorm(num_groups=num_groups, num_channels=weight.shape[0], eps=eps)
+    module.register_parameter("weight", weight)
+    module.register_parameter("bias", bias)
```
albertz, line 358: I think this is wrong. I think you cannot (should not) use `register_parameter` for a non-parameter (i.e. regular tensor).
Also, it would do the wrong thing. It is not a parameter here.
You need to also set `affine=False` in this case, and multiply by `weight`, add by `bias`.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -344,3 +344,16 @@ def norm(input: Tensor,
 
 def norm_except_dim(v: Tensor, pow: int = 2, dim: int = 0) -> Tensor:
   return modules.Norm(p=pow, axes=[i for i in range(v.dim()) if i != dim], keepdims=True)(v)
+
+
+def group_norm(input, num_groups, weight=None, bias=None, eps=1e-5) -> Tensor:
```
albertz, line 349: Types missing.

pytorch_to_returnn/torch/nn/modules/normalization.py
```diff
@@ -4,7 +4,7 @@
 import numbers
 from typing import Dict, Any, Union, List
 import tensorflow as tf
-from returnn.tf.layers.basic import LayerNormLayer
+from returnn.tf.layers.basic import LayerNormLayer, NormLayer, SubnetworkLayer
```
albertz, line 7: `SubnetworkLayer` should not be needed.

pytorch_to_returnn/torch/nn/functional.py
vieting: > You need to also set `affine=False` in this case, and multiply by `weight`, add by `bias`.

True. That's not really supported so far, right? The shapes are different and if I do something like `out = out * weight.view(1, -1, 1)` which should work in PyTorch, the new axes from `view` do not match the axes of `out` in terms of being batch dim/spatial axes in returnn.

albertz: But that should still work anyway. If not, consider this a bug, and we should fix this.

pytorch_to_returnn/torch/nn/modules/normalization.py
vieting: This happens in the case of `Fp32GroupNorm` in fairseq. Not yet exactly sure why. It is a subclass of `GroupNorm` and uses `torch.nn.functional.group_norm` in its `forward`. Should we handle that in a separate issue/PR?

pytorch_to_returnn/torch/nn/functional.py
vieting: Should we handle that in a separate issue/PR? The default use case of `GroupNorm` works because this `functional.group_norm` is not used there. However, it will currently fail for `Fp32GroupNorm` (see above).

pytorch_to_returnn/torch/nn/modules/normalization.py
```diff
@@ -53,6 +53,51 @@ def import_params_torch_to_returnn(self, *, layer: LayerNormLayer, torch_module:
...
+      self.register_parameter('bias', None)
+    self.reset_parameters()
+
+  def reset_parameters(self) -> None:
+    if self.affine:
+      init.ones_(self.weight)
+      init.zeros_(self.bias)
+
+  def extra_repr(self) -> str:
+      return '{num_groups}, {num_channels}, eps={eps}, affine={affine}'.format(**self.__dict__)
```
albertz, line 78: Remove the `extra_repr`. We currently don't use this.

pytorch_to_returnn/torch/nn/functional.py
albertz: No, the default `GroupNorm` must not be used here. In general, we must not use any module with parameters in the functional part.

albertz: But yes, I think it's easier to discuss and fix the `view` or broadcasting problem (or whatever the problem is) in another PR.

pytorch_to_returnn/torch/nn/modules/normalization.py
albertz: Ah, subclass? Can you put a link?
But anyway, that would be a separate issue. This should never be a `SubnetworkLayer`. It should be exactly be the layer you create in `create_returnn_layer_dict`.

vieting: Ok, I will create a new issue for that.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -344,3 +344,16 @@ def norm(input: Tensor,
 
 def norm_except_dim(v: Tensor, pow: int = 2, dim: int = 0) -> Tensor:
   return modules.Norm(p=pow, axes=[i for i in range(v.dim()) if i != dim], keepdims=True)(v)
+
+
+def group_norm(input: Tensor, num_groups: int, weight: Optional[Tensor] = None, bias: Optional[Tensor] = None,
+               eps: float = 1e-5) -> Tensor:
+  module = modules.GroupNorm(num_groups=num_groups, num_channels=0, eps=eps, affine=False)
+  out = module.as_returnn_torch_functional()(input)
+  assert out.shape[1] == weight.shape[0], "data should be of shape (B, F, *) and weights should be of shape (F,)"
```
albertz, line 353: Only if `weight is not None`. So move it to the branch below.
Also directly check the full `shape`. I.e.
```suggestion
  assert (out.shape[1],) == weight.shape
```
Same also for `bias`.

pytorch_to_returnn/torch/nn/functional.py
vieting: Alright!

pytorch_to_returnn/torch/nn/modules/normalization.py
```diff
@@ -53,6 +53,45 @@ def import_params_torch_to_returnn(self, *, layer: LayerNormLayer, torch_module:
...
+
+  def reset_parameters(self) -> None:
+    if self.affine:
+      init.ones_(self.weight)
+      init.zeros_(self.bias)
+
+  def import_params_torch_to_returnn(self, *, layer: NormLayer, torch_module: GroupNorm):
+    assert isinstance(layer, NormLayer)
+    session = tf.compat.v1.get_default_session()
+    layer.params["scale"].load(torch_module.weight.detach().numpy(), session=session)
```
albertz, line 80: There should be a check `if self.affine`, right?

pytorch_to_returnn/torch/nn/modules/normalization.py
```diff
@@ -53,6 +53,45 @@ def import_params_torch_to_returnn(self, *, layer: LayerNormLayer, torch_module:
...
+    layer.params["bias"].load(torch_module.bias.detach().numpy(), session=session)
+
+  def create_returnn_layer_dict(self, input: Tensor):
+    if self.num_groups == 1:
+      axes = "TF"
+    elif self.num_groups == self.num_channels:
+      axes = "T"
+    else:
+      raise NotImplementedError
```
albertz, line 89:
```suggestion
      raise NotImplementedError(f"num groups {self.num_groups}")
```

pytorch_to_returnn/torch/nn/modules/normalization.py
vieting: Absolutely

albertz (approved):

PR #24, opened 2020-12-18
add more tensor operations

vieting:
Specifically, `tensor.clone` and `tensor.device` which are ignored here as well as `tensor.pow`. Not sure if there is a more elegant way for `pow` than the `EvalLayer` but this seems pretty straight-forward.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/activation.py
+++ b/pytorch_to_returnn/torch/nn/modules/activation.py
@@ -41,6 +41,8 @@ class LogSigmoid(_ActivationReturnn):
 
 
 class Power(Module):
+  is_original_torch_module = False
+
   def __init__(self, exponent: float) -> None:
     super(Power, self).__init__()
     self.exponent = exponent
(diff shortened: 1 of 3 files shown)


pytorch_to_returnn/torch/nn/modules/activation.py
```diff
@@ -40,6 +40,17 @@ class LogSigmoid(_ActivationReturnn):
   func_name = "log_sigmoid"
 
 
+class Power(Module):
```
albertz, line 43: This is not in Torch, right?
Then you need to add this:
```
```

pytorch_to_returnn/torch/tensor.py
```diff
@@ -197,6 +203,10 @@ def sigmoid(self):
     from .nn.functional import sigmoid
     return sigmoid(self)
 
+  def pow(self, exponent):
```
albertz, line 206:
```suggestion
  def pow(self, exponent: float):
```

pytorch_to_returnn/torch/nn/modules/activation.py
vieting: Yes, right.

PR #25, opened 2020-12-18
meta infos

vieting:
I encountered some things like `torch.jit.is_scripting()`, `torch.cuda.is_available()` and `self.training` on a `Module` which I am currently just ignoring. What do you think about these?

### DIFF
--- a/pytorch_to_returnn/torch/__init__.py
+++ b/pytorch_to_returnn/torch/__init__.py
@@ -5,6 +5,6 @@ from .nn.functional import *
 from .serialization import load
 from .autograd import *
 from . import nn
-import torch.cuda
-import torch.onnx
-import torch.jit
+from . import cuda
+from . import onnx
+from . import jit
--- a/pytorch_to_returnn/torch/__init__.py
+++ b/pytorch_to_returnn/torch/__init__.py
@@ -4,7 +4,7 @@ from ._C import *
 from .nn.functional import *
 from .serialization import load
 from .autograd import *
-from .jit import *
-from .onnx import *
-from .cuda import *
 from . import nn
+import torch.cuda
+import torch.onnx
+import torch.jit
--- a/pytorch_to_returnn/torch/__init__.py
+++ b/pytorch_to_returnn/torch/__init__.py
@@ -6,4 +6,5 @@ from .serialization import load
 from .autograd import *
 from .jit import *
 from .onnx import *
+from .cuda import *
 from . import nn
--- a/pytorch_to_returnn/torch/cuda.py
+++ b/pytorch_to_returnn/torch/cuda.py
@@ -1,8 +1,2 @@
-def is_available():
-  """
-  Always returns False for now
-
-  :return: False
-  :rtupe: bool
-  """
-  return False
+def is_available() -> bool:
+  return False  # stub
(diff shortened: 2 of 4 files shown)


pytorch_to_returnn/torch/cuda.py
```diff
@@ -0,0 +1,8 @@
+def is_available():
```
albertz, line 1:
```suggestion
def is_available() -> bool:
```

pytorch_to_returnn/torch/cuda.py
```diff
@@ -0,0 +1,8 @@
+def is_available():
+  """
+  Always returns False for now
+
+  :return: False
+  :rtupe: bool
+  """
```
albertz, line 7: Bad documentation.
Typo.
You are redundant. Use `:return:` to describe what you return, but then don't write it again twice as the description.

But anyway, this is wrong. We probably will change this. You should not document as if that is how it should be. If you really would want a documentation, write sth like "Dummy/stub. Not implemented yet...". Or just leave it away.

```suggestion
```

pytorch_to_returnn/torch/jit.py
```diff
@@ -6,3 +6,14 @@ def export(fn):
...
+
+def is_scripting():
+  """
+  Always returns False as we never are in script mode
+  https://pytorch.org/docs/master/jit_language_reference.html#torch.jit.is_scripting
+
+  :return: False
+  :rtupe: bool
+  """
+  return False
```
albertz, line 19:
```suggestion
def is_scripting() -> bool:
  return False  # stub
```

pytorch_to_returnn/torch/__init__.py
```diff
@@ -6,4 +6,5 @@
 from .autograd import *
 from .jit import *
 from .onnx import *
+from .cuda import *
```
albertz, line 9: Why do you do this? Is this in the original Torch as well? I.e. there is a global function `torch.is_available`? That doesn't really make sense...

vieting: That's true, doesn't make sense. I'll remove it.

vieting: Isn't it the same for `jit`? Can we just remove it or should it instead be `import torch.cuda` and `import torch.jit`?

albertz: Probably. I haven't checked it.
In any case, it should be just the same as in the original PyTorch.

pytorch_to_returnn/torch/__init__.py
```diff
@@ -4,6 +4,7 @@
 from .nn.functional import *
 from .serialization import load
 from .autograd import *
-from .jit import *
-from .onnx import *
 from . import nn
+import torch.cuda
+import torch.onnx
+import torch.jit
```
albertz, line 10: Why do you do this?
Is this done in PyTorch?

albertz: Also, this is wrong in any case.

albertz, line 10:
```suggestion
from . import cuda
from . import onnx
from . import jit
```
But double check that this is also done in PyTorch. Otherwise it should be left away.

vieting: This is done in PyTorch, see https://github.com/pytorch/pytorch/blob/master/torch/__init__.py#L570
That's why I took it over.

PR #27, opened 2020-12-18
subnetwork problem in group norm

vieting:
I slightly modified the `Fp32GroupNorm` test case to reproduce the problem that I mentioned earlier in #18. If `Fp32GroupNorm` is wrapped inside a `Sequential`, the layer that is input to `GroupNorm.import_params_torch_to_returnn` will be a `SubnetworkLayer`. Do you have any ideas on this?


albertz:
You should not modify an existing test case to test for sth different. Make a new test case.

vieting:
I thought it would be easier to see the detail that differed but sure, it's better to have a new test case.

PR #28, opened 2020-12-22
add linear functional

vieting:
Add `torch.nn.functional.linear` using the new module `FunctionalLinear`. This first needs the corresponding PR in returnn https://github.com/rwth-i6/returnn/pull/416

### DIFF
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -282,8 +282,10 @@ def max(*inputs: Tensor) -> Tensor:
 
 
 def linear(input: Tensor, weight: Tensor, bias: Optional[Tensor] = None):
-  mod = modules.FunctionalLinear()
-  return mod(input, weight, bias)
+  output = input.matmul(weight.t())
+  if bias is not None:
+    output += bias
+  return output
 
 
 def conv1d(
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -117,7 +117,7 @@ def mul(x: Tensor, y: Tensor) -> Tensor:
 
 def matmul(input: Tensor, other: Tensor, *, out: Optional[Tensor] = None) -> Tensor:
   assert out is None, "not implemented otherwise"
-  mod = modules.DotLayer()
+  mod = modules.Matmul()
   if len(input.shape) > len(other.shape):
     other = other.expand(list(input.shape[:-len(other.shape)]) + [-1] * len(other.shape))
   elif len(input.shape) < len(other.shape):
@@ -218,6 +218,7 @@ def t(input: Tensor):
   elif len(input.shape) == 2:
     return transpose(input, 0, 1)
   else:
+    # https://pytorch.org/docs/stable/generated/torch.t.html#torch.t
     raise ValueError("t() expects input to be <= 2-D tensor")
 
 
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -212,6 +212,15 @@ def transpose(input: Tensor, dim0: int, dim1: int):
   return tensorflow_transpose(input, perm={dim0: dim1, dim1: dim0})
 
 
+def t(input: Tensor):
+  if len(input.shape) < 2:
+    return input
+  elif len(input.shape) == 2:
+    return transpose(input, 0, 1)
+  else:
+    raise ValueError("t() expects input to be <= 2-D tensor")
+
+
 def tensorflow_transpose(input: Tensor, perm: Optional[Union[Dict[int, int], Tuple[int, ...], List[int]]]):
   """
   Note: This function is added by us, not available in original PyTorch.
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -115,6 +115,16 @@ def mul(x: Tensor, y: Tensor) -> Tensor:
   return modules.BinaryOperator(kind="mul")(cast(x, dtype), cast(y, dtype))
 
 
+def matmul(input: Tensor, other: Tensor, *, out: Optional[Tensor] = None) -> Tensor:
+  assert out is None, "not implemented otherwise"
+  mod = modules.DotLayer()
+  if len(input.shape) > len(other.shape):
+    other = other.expand(list(input.shape[:-len(other.shape)]) + [-1] * len(other.shape))
+  elif len(input.shape) < len(other.shape):
+    input = input.expand(list(other.shape[:-len(input.shape)]) + [-1] * len(input.shape))
+  return mod(input, other)
+
+
 def truediv(x: Tensor, y: Tensor) -> Tensor:
   dtype = result_type(x, y)
   return modules.BinaryOperator(kind="truediv")(cast(x, dtype), cast(y, dtype))
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -262,6 +262,11 @@ def max(*inputs: Tensor) -> Tensor:
   return modules.Max()(*inputs)
 
 
+def linear(input: Tensor, weight: Tensor, bias: Optional[Tensor] = None):
+  mod = modules.FunctionalLinear()
+  return mod(input, weight, bias)
+
+
 def conv1d(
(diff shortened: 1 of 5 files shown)


pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -60,6 +60,27 @@ def import_params_torch_to_returnn(self, *, layer: LinearLayer, torch_module: Li
...
+    if bias is not None:
+      assert bias.shape == (out_features,)
+    return {
+      "class": "linear", "from": self._get_input_layer_name(input),
+      "n_out": out_features,
+      "with_bias": bias is not None,
+      "bias": self._get_input_layer_name(bias) if bias is not None else None,
+      "weights": self._get_input_layer_name(weight),
+      "use_transposed_weights": True,
+      "activation": None}
```
albertz, line 81: Rather than defining `create_returnn_layer_dict`, you should define `forward` and use `functional.matmul` and just `+` for the bias.

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -60,6 +60,27 @@ def import_params_torch_to_returnn(self, *, layer: LinearLayer, torch_module: Li
       layer.params["b"].load(torch_module.bias.detach().numpy(), session=session)
 
 
+class FunctionalLinear(Module):
+  is_original_torch_module = False
+
+  def __init__(self):
+    super(FunctionalLinear, self).__init__()
+
```
albertz, line 68:
```suggestion
```

vieting: Is there a difference?

pytorch_to_returnn/torch/nn/modules/linear.py
vieting: True, but then actually we don't need this module at all, right? We can just do it in `functional.linear`.

pytorch_to_returnn/torch/nn/modules/linear.py
albertz: Functionality-wise, no, but it's cleaner.

pytorch_to_returnn/torch/nn/modules/linear.py
albertz: Yes. This is maybe a matter of taste. In one case, it will end up in a subnetwork, in the other not.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -197,6 +207,15 @@ def transpose(input: Tensor, dim0: int, dim1: int):
   return tensorflow_transpose(input, perm={dim0: dim1, dim1: dim0})
 
 
+def t(input: Tensor):
+  if len(input.shape) < 2:
+    return input
+  elif len(input.shape) == 2:
+    return transpose(input, 0, 1)
+  else:
+    raise ValueError("t() expects input to be <= 2-D tensor")
```
albertz, line 216: Is that the case? Or not implemented?

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -60,6 +60,42 @@ def import_params_torch_to_returnn(self, *, layer: LinearLayer, torch_module: Li
...
+      "class": "linear", "from": self._get_input_layer_name(input),
+      "n_out": out_features,
+      "with_bias": bias is not None,
+      "bias": self._get_input_layer_name(bias) if bias is not None else None,
+      "weights": self._get_input_layer_name(weight),
+      "use_transposed_weights": True,
+      "activation": None}
+
+
+class DotLayer(Module):
```
albertz, line 84: I wonder whether we maybe should call it `Matmul` instead.

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -60,6 +60,42 @@ def import_params_torch_to_returnn(self, *, layer: LinearLayer, torch_module: Li
...
+  """
+  is_original_torch_module = False
+
+  def __init__(self):
+    super(DotLayer, self).__init__()
+
+  def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]:
+    sources = [self._get_input_layer_name(source) for source in inputs]
+    assert len(sources) == 2
+    return {"class": "dot", "from": sources}
```
albertz, line 96: This is probably incomplete, right? You definitely need to specify the axes.

pytorch_to_returnn/torch/nn/functional.py
vieting: Yes, that's stated in the PyTorch docs https://pytorch.org/docs/stable/generated/torch.t.html#torch.t

pytorch_to_returnn/torch/nn/modules/linear.py
vieting: Yes, I guess that fits nicer into the PyTorch naming

pytorch_to_returnn/torch/nn/functional.py
albertz: Ok. Maybe add a small comment. This is a bit unintuitive (because NumyPy and TensorFlow allow any N-D I think).

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -93,7 +93,11 @@ def __init__(self):
   def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]:
     sources = [self._get_input_layer_name(source) for source in inputs]
     assert len(sources) == 2
-    return {"class": "dot", "from": sources}
+    red1 = inputs[0].returnn_naming_entry.get_returnn_axis_description(-1)
+    red2 = inputs[1].returnn_naming_entry.get_returnn_axis_description(-2)
+    var1 = inputs[0].returnn_naming_entry.get_returnn_axis_description(-2)
+    var2 = inputs[1].returnn_naming_entry.get_returnn_axis_description(-1)
```
albertz, line 99: Better make this consistent as it is done in other layers. Via `self.get_axis...`

vieting: I'm not sure what you mean here. I don't see modules which use a function `self.get_axis...`

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -60,6 +60,59 @@ def import_params_torch_to_returnn(self, *, layer: LinearLayer, torch_module: Li
...
+  def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]:
+    sources = [self._get_input_layer_name(source) for source in inputs]
+    assert len(sources) == 2
+
+    assert len(inputs[0].shape) >= 2, "not implemented otherwise"
+    assert len(inputs[1].shape) >= 2, "not implemented otherwise"
+    red1 = inputs[0].returnn_naming_entry.get_returnn_axis_description(-1)
+    red2 = inputs[1].returnn_naming_entry.get_returnn_axis_description(-2)
+    var1 = [inputs[0].returnn_naming_entry.get_returnn_axis_description(-2)]
+    var2 = [inputs[1].returnn_naming_entry.get_returnn_axis_description(-1)]
```
albertz, line 96: See other layers. Make this consistent. You can use `self._get_axis...` (or sth like that). Do not access `returnn_naming_entry`.

pytorch_to_returnn/torch/nn/modules/linear.py
albertz: This still should be cleaned up. I think you don't need it at all anymore, right?

pytorch_to_returnn/torch/nn/modules/linear.py
albertz: See how all other layers get the axis.

pytorch_to_returnn/torch/nn/modules/linear.py
albertz: `self._get_input_axis_to_returnn`

pytorch_to_returnn/torch/nn/modules/linear.py
albertz: `self._get_input_axis_to_returnn`

pytorch_to_returnn/torch/nn/modules/linear.py
vieting: That's right.

albertz (approved):

PR #30, opened 2021-02-11
Add torch.cat for the case that concatenation happens in the feature dim

lucasoethe:
In case concatenation of tensors happens in the feature dimension we can take advantage of the fact that the RETURNN copy layer concatenates it's inputs in the feature dimension.

### DIFF
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -11,6 +11,7 @@ from typing import Optional, Union, List, Tuple, Dict, TypeVar, Sequence
 from . import modules
 from ..tensor import Tensor
 from .._C import Size, dtype as _dtype
+from ...naming import Naming
 
 
 _number = Union[int, float, numpy.ndarray, numpy.number]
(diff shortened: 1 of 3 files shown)


pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -10,13 +10,8 @@
 class Copy(Module):
   is_original_torch_module = False
 
-  def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
-    return {"class": "copy", "from": self._get_input_layer_name(input)}
-
-  def make_output_tensor_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase) -> Tensor:
-    assert len(inputs_flat) == 1
-    return inputs_flat[0]
-
```
albertz, line 19: I would leave that in.

Make a new separate module `Cat`.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -46,6 +47,14 @@ def cast(input: Union[_T, Tensor, _number], dtype: Union[str, _dtype]) -> Union[
     return input
   return modules.Cast(dtype=dtype)(input)
 
+def cat(tensors, dim=0):
+  from .modules.operator import Copy
+  naming = Naming.get_instance()
+  for tensor in tensors:
+    returnn_data = naming.tensors[tensor].returnn_data
+    assert returnn_data.get_axis_from_description(dim) == returnn_data.feature_dim_axis, "Concatenation in dimensions other than the feature dimension is currently not supported."
```
albertz, line 55: This is wrong, in multiple ways.

`get_axis_from_description` doesn't make any sense here. `dim` is a Torch dimension (integer). `get_axis_from_description` is for RETURNN axes descriptions (usually string).

Anyway, move this logic to the `Cat` module (see my other comment). And in there, handle the axis mapping (Torch -> RETURNN) as we do in the other layers. And then you can check that the RETURNN axis (int) matches the feature dim axis for all the given tensors, and if not, fail with some exception.

albertz (approved):

PR #33, opened 2021-04-21
allow more general reshapes

vieting:
Generalize `reshape` to allow reshapes of `(..., a*b, c,...) -> (..., a, b*c,...)` or `(..., a, b*c,...) -> (..., a*b, c,...)`.

### DIFF
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -224,22 +224,21 @@ def test_t():
 
 
 def test_reshape():
-  n_batch, n_feature, n_time = 3, 8, 5
+  n_batch, n_time, n_feature_1, n_feature_2 = 3, 5, 8, 6
 
   def model_func_1(wrapped_import, inputs: torch.Tensor):
     # test case (..., a*b, c,...) -> (..., a, b*c,...)
-    return inputs.view(n_batch, n_feature // 2, n_time * 2)
+    return inputs.view(n_batch, n_time, n_feature_1 // 2, n_feature_2 * 2)
 
   def model_func_2(wrapped_import, inputs: torch.Tensor):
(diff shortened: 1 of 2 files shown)


tests/test_layers.py
```diff
@@ -223,6 +223,25 @@ def model_func(wrapped_import, inputs: torch.Tensor):
   verify_torch_and_convert_to_returnn(model_func, inputs=x)
 
 
+def test_reshape():
+  n_batch, n_feature, n_time = 3, 8, 5
+
+  def model_func_1(wrapped_import, inputs: torch.Tensor):
+    # test case (..., a*b, c,...) -> (..., a, b*c,...)
+    return inputs.view(n_batch, n_feature // 2, n_time * 2)
```
albertz, line 231: This is not exactly correct for the time dim. On RETURNN side, `inputs` has shape [B,D,T] where T is dynamic. Then you make it static here. This might break at some point. I actually wonder that it works without error now.

I would introduce another second static axis for `inputs`, and do this reshape test on two static axes (not involving time or batch).

When you want to test specifically for correct handling of time or batch, make a separate test for it (but maybe also separate PR, so we can already merge this, and then do further work on it).

tests/test_layers.py
```diff
@@ -223,6 +223,25 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
+
+  def model_func_2(wrapped_import, inputs: torch.Tensor):
+    # test case (..., a, b*c,...) -> (..., a*b, c,...)
+    return inputs.view(n_batch, n_time * 2, n_feature // 2)
```
albertz, line 235: Same here.

albertz (approved):

PR #34, opened 2021-04-21
add torch.bmm

vieting:
Add batch matix-matrix product `torch.bmm` using existing `matmul`.

### DIFF
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -202,6 +202,25 @@ def test_matmul():
   verify_torch_and_convert_to_returnn(model_func, inputs=x)
 
 
+def test_bmm():
+  n_in, n_out = 11, 13
+  n_batch, n_time = 3, 5
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    if typing.TYPE_CHECKING or not wrapped_import:
+      import torch
+    else:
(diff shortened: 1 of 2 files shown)


tests/test_layers.py
```diff
@@ -202,6 +202,25 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
+  n_in, n_out = 11, 13
+  n_batch, n_time = 3, 5
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    if typing.TYPE_CHECKING or not wrapped_import:
+      import torch
+    else:
+      torch = wrapped_import("torch")
+
+    x = inputs.new_zeros(inputs.shape[0], inputs.shape[2], n_out) + 1  # (T, F_in, F_out)
```
albertz, line 215: Make this test simpler. Use the batch-dim. E.g. like:
```suggestion
    inputs = inputs.transpose(0, 1)
    x = inputs.new_zeros(inputs.shape[0], inputs.shape[2], n_out) + 1  # (T, F_in, F_out)
```
Handling the time-dim properly here is a separate issue, which we can fix separately. It's not so much about `bmm`.

albertz (approved):

PR #36, opened 2021-04-26
fix for spatial axes with same dim tag

vieting:
Having tensors where multiple axes have the same dim tag lead to problems in `_get_input_axis_to_returnn` as well as when creating `out_returnn_axis_to_torch_axis` in `_base_get_output_shape_from_returnn`. Using the new dim tag format `stag-single:<idx>:<name>` for spatial axes in returnn (see https://github.com/rwth-i6/returnn/pull/490), we can circumvent this.

### DIFF
--- a/pytorch_to_returnn/naming/tensor.py
+++ b/pytorch_to_returnn/naming/tensor.py
@@ -108,7 +108,7 @@ class TensorEntry:
       if len(axes_with_same_dim_tag) == 1:
         return f"stag:{dim_tag.description}"
       else:
-        return f"stag-single:{axis - ndim}:{dim_tag.description}"
+        return f"stag-single:{axes_with_same_dim_tag.index(axis) - len(axes_with_same_dim_tag)}:{dim_tag.description}"
     static_axes = self.returnn_data.get_static_axes()
     if axis in static_axes:
       return f"static:{static_axes.index(axis)}"
--- a/pytorch_to_returnn/naming/tensor.py
+++ b/pytorch_to_returnn/naming/tensor.py
@@ -104,7 +104,11 @@ class TensorEntry:
     dim_tag = self.returnn_data.get_dim_tag(axis)
     assert dim_tag.kind == DimensionTag.Types.Spatial
     if dim_tag.dyn_size is not None:
-      return f"stag:{dim_tag.description}"
+      axes_with_same_dim_tag = [ax for ax in range(ndim) if dim_tag == self.returnn_data.get_dim_tag(ax)]
+      if len(axes_with_same_dim_tag) == 1:
+        return f"stag:{dim_tag.description}"
+      else:
+        return f"stag-single:{axis - ndim}:{dim_tag.description}"
     static_axes = self.returnn_data.get_static_axes()
     if axis in static_axes:
       return f"static:{static_axes.index(axis)}"
--- a/pytorch_to_returnn/torch/nn/modules/module.py
+++ b/pytorch_to_returnn/torch/nn/modules/module.py
@@ -13,7 +13,7 @@ from ...autograd import no_grad
 from ...utils.hooks import RemovableHandle
 from ....naming import Naming, CallEntry, TensorEntry
 from returnn.tf.layers.basic import LayerBase, SubnetworkLayer
-from returnn.tf.util.data import Data
+from returnn.tf.util.data import Data, DimensionTag
 
 # See https://mypy.readthedocs.io/en/latest/generics.html#generic-methods-and-generic-self for the use
 # of `T` to annotate `self`. Many methods of `Module` return `self` and we want those return values to be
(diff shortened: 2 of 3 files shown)


pytorch_to_returnn/naming/tensor.py
```diff
@@ -104,7 +104,11 @@ def get_returnn_axis_description(self, torch_axis: int) -> str:
     dim_tag = self.returnn_data.get_dim_tag(axis)
     assert dim_tag.kind == DimensionTag.Types.Spatial
     if dim_tag.dyn_size is not None:
-      return f"stag:{dim_tag.description}"
+      axes_with_same_dim_tag = [ax for ax in range(ndim) if dim_tag == self.returnn_data.get_dim_tag(ax)]
+      if len(axes_with_same_dim_tag) == 1:
+        return f"stag:{dim_tag.description}"
+      else:
+        return f"stag-single:{axis - ndim}:{dim_tag.description}"
```
albertz, line 111: The index calculation looks wrong.
```suggestion
        return f"stag-single:{axes_with_same_dim_tag.index(axis) - len(axes_with_same_dim_tag)}:{dim_tag.description}"
```

pytorch_to_returnn/torch/nn/modules/module.py
```diff
@@ -757,6 +757,11 @@ def _base_get_output_shape_from_returnn(inputs_flat: List[Tensor], layer: LayerB
     # DimensionTag.get_all_dimension_tags, etc.
     # However, this is simpler, and also more consistent with get_returnn_axis_description.
 
+    def _get_unique_spatial_dim_tag_name(data, axis):
```
albertz, line 760: Specify types:
```suggestion
    def _get_unique_spatial_dim_tag_name(data: Data, axis: int) -> str:
```

pytorch_to_returnn/torch/nn/modules/module.py
```diff
@@ -773,7 +778,7 @@ def _base_get_output_shape_from_returnn(inputs_flat: List[Tensor], layer: LayerB
       if x.returnn_data.have_batch_axis():
         batch_size = input.shape[x.returnn_axis_from_torch_axis[x.returnn_data.batch_dim_axis]]
       for i in x.returnn_data.get_dynamic_axes():
-        dim_tag = x.returnn_data.get_dim_tag(i)
+        dim_tag = _get_unique_spatial_dim_tag_name(x.returnn_data, i)
```
albertz, line 781: I'm not sure if `dim_tag_name` or so is maybe better.

But in any case, you need to fix the documentation of `dyn_size_dim_tag_to_spatial_idx_and_torch_dim`. Now it's not anymore "RETURNN dim tag" but "RETURNN unique dim tag name" or so.

pytorch_to_returnn/torch/nn/modules/module.py
```diff
@@ -757,6 +757,11 @@ def _base_get_output_shape_from_returnn(inputs_flat: List[Tensor], layer: LayerB
     # DimensionTag.get_all_dimension_tags, etc.
     # However, this is simpler, and also more consistent with get_returnn_axis_description.
 
+    def _get_unique_spatial_dim_tag_name(data, axis):
+      dim_tag = data.get_dim_tag(axis)
+      idx = data.get_axes_by_tag_name(dim_tag.description).index(axis)
+      return f"stag-single:{idx}:{dim_tag.description}"
```
albertz, line 763: Why do you need this as a string here?
I think better would be:
```suggestion
      return dim_tag, idx
```

albertz: And then change types, function name, and other documentation accordingly.

albertz: E.g. the function could be named `_get_spatial_dim_tag_single` (to be consistent to "stag-single").

vieting: It does not need to be a string, but I somehow need the index to keep the `dim_tag`s apart. I could also use a tuple `(dim_tag, idx)` as key to `dyn_size_dim_tag_to_spatial_idx_and_torch_dim` (which could possibly renamed). Do you have preferances here?

albertz: But this is exactly what I suggested? See my suggested change above.

vieting: Oh sorry, somehow the two additional comments did not show up before.

pytorch_to_returnn/torch/nn/modules/module.py
```diff
@@ -773,12 +779,12 @@ def _base_get_output_shape_from_returnn(inputs_flat: List[Tensor], layer: LayerB
       if x.returnn_data.have_batch_axis():
         batch_size = input.shape[x.returnn_axis_from_torch_axis[x.returnn_data.batch_dim_axis]]
       for i in x.returnn_data.get_dynamic_axes():
-        dim_tag = x.returnn_data.get_dim_tag(i)
+        dim_tag, idx = _get_spatial_dim_tag_and_single_index(x.returnn_data, i)
```
albertz, line 782:
```suggestion
        dim_tag_ext = _get_spatial_dim_tag_and_single_index(x.returnn_data, i)
```
No need to separate them both here. That makes it also simpler when we need to change/extend `_get_spatial_dim_tag_and_single_index` at some point.

albertz (approved):

PR #38, opened 2021-04-28
fix loading params to returnn

vieting:
Loading the params to returnn works correctly when having a module `mod = torch.nn.Linear(n_in, n_out)` and calling it directly, i.e., `y = mod(x)`. However, if the params are used directly like `mod.weight`, they are not loaded in returnn currently.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/variable.py
+++ b/pytorch_to_returnn/torch/nn/modules/variable.py
@@ -26,11 +26,12 @@ class Variable(Module):
     naming = Naming.get_instance()
     values = None
     if naming.import_params_from_torch_namespace and self.parent_mod:
-      parent_mod, param_name = self.parent_mod
-      mod_abs_name = naming.get_module_abs_id_name(parent_mod)
-      torch_mod = naming.import_params_from_torch_namespace.get_module_by_abs_id_name(mod_abs_name)
-      torch_param = getattr(torch_mod, param_name)
-      values = torch_param.detach().numpy()
+      if not ".tmp_root" in layer.network.name:  # temp layer
+        parent_mod, param_name = self.parent_mod
+        mod_abs_name = naming.get_module_abs_id_name(parent_mod)
+        torch_mod = naming.import_params_from_torch_namespace.get_module_by_abs_id_name(mod_abs_name)
+        torch_param = getattr(torch_mod, param_name)
+        values = torch_param.detach().numpy()
     if values is None:
       values = self.param.detach().numpy()
     assert isinstance(layer, VariableLayer)
--- a/pytorch_to_returnn/torch/nn/modules/variable.py
+++ b/pytorch_to_returnn/torch/nn/modules/variable.py
@@ -26,7 +26,7 @@ class Variable(Module):
     naming = Naming.get_instance()
     values = None
     if naming.import_params_from_torch_namespace and self.parent_mod:
-      if ".tmp_root" not in layer.network.name:  # temp layer
+      if not layer.get_absolute_name().startswith("."):  # temp layer
         parent_mod, param_name = self.parent_mod
         mod_abs_name = naming.get_module_abs_id_name(parent_mod)
         torch_mod = naming.import_params_from_torch_namespace.get_module_by_abs_id_name(mod_abs_name)
--- a/pytorch_to_returnn/torch/nn/modules/variable.py
+++ b/pytorch_to_returnn/torch/nn/modules/variable.py
@@ -26,7 +26,7 @@ class Variable(Module):
     naming = Naming.get_instance()
     values = None
     if naming.import_params_from_torch_namespace and self.parent_mod:
-      if not ".tmp_root" in layer.network.name:  # temp layer
+      if ".tmp_root" not in layer.network.name:  # temp layer
         parent_mod, param_name = self.parent_mod
         mod_abs_name = naming.get_module_abs_id_name(parent_mod)
         torch_mod = naming.import_params_from_torch_namespace.get_module_by_abs_id_name(mod_abs_name)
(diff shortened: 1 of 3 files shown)


pytorch_to_returnn/torch/nn/modules/variable.py
```diff
@@ -17,16 +17,25 @@ def __init__(self, param: Parameter):
     super(Variable, self).__init__()
     assert isinstance(param, Parameter)
     self.param = param
+    self.parent_mod = None
```
albertz, line 20: Type should be documented.
```suggestion
    self.parent_mod = None  # type: Optional[Tuple[Module, str]]
```
But maybe this can directly be a param in `__init__`? That would be cleaner instead of setting it from outside.

pytorch_to_returnn/torch/nn/modules/variable.py
```diff
@@ -17,16 +17,25 @@ def __init__(self, param: Parameter):
...
 
   def create_returnn_layer_dict(self, *inputs):  # ignore inputs
     return {"class": "variable", "add_batch_axis": False, "shape": self.param.shape}
 
   def make_output_tensor_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase) -> Tensor:
+    naming = Naming.get_instance()
+    values = self.param.detach().numpy()
```
albertz, line 27: Better set `values = None` here, and then below sth like `if values is None: values = self.param.detach().numpy()`.

pytorch_to_returnn/torch/nn/modules/variable.py
```diff
@@ -17,16 +17,25 @@ def __init__(self, param: Parameter):
...
+    if naming.import_params_from_torch_namespace and self.parent_mod:
+      if self.parent_mod[0].is_original_torch_module:
+        mod_abs_name = naming.get_module_abs_id_name(self.parent_mod[0])
+        torch_mod = naming.import_params_from_torch_namespace.get_module_by_abs_id_name(mod_abs_name)
+        values = torch_mod.weight.detach().numpy()
```
albertz, line 32:
```suggestion
        torch_param = getattr(torch_mod, param_name)
        values = torch_param.detach().numpy()
```

pytorch_to_returnn/torch/nn/modules/variable.py
```diff
@@ -17,16 +17,25 @@ def __init__(self, param: Parameter):
...
+    self.parent_mod = None
 
   def create_returnn_layer_dict(self, *inputs):  # ignore inputs
     return {"class": "variable", "add_batch_axis": False, "shape": self.param.shape}
 
   def make_output_tensor_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase) -> Tensor:
+    naming = Naming.get_instance()
+    values = self.param.detach().numpy()
+    if naming.import_params_from_torch_namespace and self.parent_mod:
+      if self.parent_mod[0].is_original_torch_module:
```
albertz, line 29: Actually I'm not sure anymore if we really need this check here.

pytorch_to_returnn/torch/nn/modules/variable.py
vieting: Yes, I added in as an argument to `__init__`.

pytorch_to_returnn/torch/nn/modules/variable.py
vieting: Me neither. Since there is no clear argument for checking it, I removed it for. Or would you prefer to keep it?

albertz: No, remove it.

vieting:
We have to check whether the layer is a temp layer, similar to `apply_call` https://github.com/rwth-i6/pytorch-to-returnn/blob/main/pytorch_to_returnn/naming/call.py#L160

However, the network name is `root/.tmp_root...`, so I modified the check. Actually, in `apply_call` the name is also `root/.tmp_root...` so the check there never applies, at least for none of the tests in `test_layers.py`. Is that intended? What do you think?

pytorch_to_returnn/torch/nn/modules/variable.py
```diff
@@ -26,11 +26,12 @@ def make_output_tensor_from_returnn(self, inputs_flat: List[Tensor], layer: Laye
     naming = Naming.get_instance()
     values = None
     if naming.import_params_from_torch_namespace and self.parent_mod:
-      parent_mod, param_name = self.parent_mod
-      mod_abs_name = naming.get_module_abs_id_name(parent_mod)
-      torch_mod = naming.import_params_from_torch_namespace.get_module_by_abs_id_name(mod_abs_name)
-      torch_param = getattr(torch_mod, param_name)
-      values = torch_param.detach().numpy()
+      if not ".tmp_root" in layer.network.name:  # temp layer
```
albertz, line 29: You always should use `if ... not in ...` (PyCharm should tell you, I think).

But I also think there should be a better way to check for a temp layer. I'm not sure you really should look at `layer.network.name` here.

vieting: We could do `if naming.tmp_eager_root_namespace.name not in layer.network.name:` in order not to hard-code `".tmp_root"`. For `layer.network.name` I currently don't know an alternative. I'm open to suggestions of course. Might also be a separate issue?

albertz: I just checked the code now. It's also wrong in `apply_call`. It's wrong because the meaning of `network.name` changed (it was never really meant to represent anything except for debugging).
We should use `layer.get_absolute_name()` (both in `apply_call` and here). I just pushed that for `apply_call` in 306557fe17b00ea3acfd4.

vieting: Ah yes, good!

albertz (approved):

PR #39, opened 2021-04-30
add F.multi_head_attention_forward

vieting:
Adds `F.multi_head_attention_forward` and a test case which is a simplified version of the `MultiheadAttention` module in fairseq.

### DIFF
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -651,23 +651,24 @@ def multi_head_attention_forward(
     assert bias_k is None
     assert bias_v is None
 
-  q = q.contiguous().view(tgt_len, bsz * num_heads, head_dim).transpose(0, 1)
+  # reshape to (bsz, num_heads, tgt_len, head_dim)
+  q = q.contiguous().view(tgt_len, bsz, num_heads, head_dim).transpose(0, 1).transpose(1, 2)
   if k is not None:
-    k = k.contiguous().view(-1, bsz * num_heads, head_dim).transpose(0, 1)
+    k = k.contiguous().view(-1, bsz, num_heads, head_dim).transpose(0, 1).transpose(1, 2)
   if v is not None:
-    v = v.contiguous().view(-1, bsz * num_heads, head_dim).transpose(0, 1)
+    v = v.contiguous().view(-1, bsz, num_heads, head_dim).transpose(0, 1).transpose(1, 2)
 
   if static_k is not None:
     assert static_k.size(0) == bsz * num_heads
     assert static_k.size(2) == head_dim
-    k = static_k
+    k = static_k.view(bsz, num_heads, -1, head_dim)
 
   if static_v is not None:
     assert static_v.size(0) == bsz * num_heads
     assert static_v.size(2) == head_dim
-    v = static_v
+    v = static_v.view(bsz, num_heads, -1, head_dim)
 
-  src_len = k.size(1)
+  src_len = k.size(2)
 
   if key_padding_mask is not None:
     assert key_padding_mask.size(0) == bsz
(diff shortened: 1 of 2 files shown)


pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -434,3 +435,281 @@ def dropout(input: Tensor, p: float = 0.5, training: bool = True, inplace: bool
     return input
   if p > 0.0:
     return modules.Dropout(p=p, inplace=inplace).as_returnn_torch_functional()(input)
+
```
albertz, line 438: One more empty line.
```suggestion


```

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -434,3 +435,281 @@ def dropout(input: Tensor, p: float = 0.5, training: bool = True, inplace: bool
...
+      - static_k: :math:`(N*num_heads, S, E/num_heads)`, where S is the source sequence length,
+        N is the batch size, E is the embedding dimension. E/num_heads is the head dimension.
+      - static_v: :math:`(N*num_heads, S, E/num_heads)`, where S is the source sequence length,
+        N is the batch size, E is the embedding dimension. E/num_heads is the head dimension.
+      Outputs:
+      - attn_output: :math:`(L, N, E)` where L is the target sequence length, N is the batch size,
+        E is the embedding dimension.
+      - attn_output_weights: :math:`(N, L, S)` where N is the batch size,
+        L is the target sequence length, S is the source sequence length.
+  """
```
albertz, line 516: Is this exactly the original implementation? Or did you change it somewhere? You should document/comment that.

Also, I avoided to copy all the exact documentation from PyTorch, as I see PyTorch itself as the documentation. We should stay consistent here. (If you think this is not a good idea in general, maybe open an issue, and let's discuss there. I'm not really sure about it.)

albertz:
This actually doesn't fix anything, except that it adds `multi_head_attention_forward`. So all worked fine now? (After the previous fixes.)

pytorch_to_returnn/torch/nn/functional.py
vieting: I copied the exact implementation but then modified some parts. You can see it here from the commits, but I will make it more clear with comments as well.

I'm fine with removing the documentation

vieting:
Yes, all the fixes that I needed to get this to work were handled in separate PRs before. Not sure what you mean with *all* worked fine now, but at least the test here works. It's a rather simple case though, I didn't test all the possibilities like e.g. using `bias_k`, `bias_v`, `add_zero_attn`.

albertz:
> but at least the test here works

Yes, that's what I mean. I assume it reflects what you actually need to go forward with Wav2Vec2 or other things.

pytorch_to_returnn/torch/nn/functional.py
albertz: > I copied the exact implementation but then modified some parts. You can see it here from the commits, but I will make it more clear with comments as well.

I will stash the commits here (because here is not too much value in keeping them separate).
But in any case, it would be helpful if this is documented (very shortly, just a single comment on some changed part).

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -646,7 +617,7 @@ def multi_head_attention_forward(
       or attn_mask.dtype == torch.bool
     ), "Only float, byte, and bool types are supported for attn_mask, not {}".format(attn_mask.dtype)
     if attn_mask.dtype == torch.uint8:
-      warnings.warn("Byte tensor for attn_mask in nn.MultiheadAttention is deprecated. Use bool tensor instead.")
+      print("Warning: Byte tensor for attn_mask in nn.MultiheadAttention is deprecated. Use bool tensor instead.")
```
albertz, line 620: Why this change? Shouldn't we also use `logging`? (That's what PyTorch `warnings` is, I assume.)

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -651,60 +651,60 @@ def multi_head_attention_forward(
     assert bias_k is None
     assert bias_v is None
 
-  q = q.contiguous().view(tgt_len, bsz * num_heads, head_dim).transpose(0, 1)
+  # reshape to (bsz, num_heads, tgt_len, head_dim)
+  q = q.contiguous().view(tgt_len, bsz, num_heads, head_dim).transpose(0, 1).transpose(1, 2)
   if k is not None:
-    k = k.contiguous().view(-1, bsz * num_heads, head_dim).transpose(0, 1)
+    k = k.contiguous().view(-1, bsz, num_heads, head_dim).transpose(0, 1).transpose(1, 2)
```
albertz, line 657: Is this not the same as `...transpose(0, 1, 2)` or sth like that? (I'm pretty sure that you can combine it to a single transpose.)

vieting: It's possible via `permute` but that's not implemented yet and I had some strange issues when I wanted to add it so I used this instead. Of course it would be nicer with `permute` though.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -651,60 +651,60 @@ def multi_head_attention_forward(
...
   if v is not None:
-    v = v.contiguous().view(-1, bsz * num_heads, head_dim).transpose(0, 1)
+    v = v.contiguous().view(-1, bsz, num_heads, head_dim).transpose(0, 1).transpose(1, 2)
 
   if static_k is not None:
     assert static_k.size(0) == bsz * num_heads
     assert static_k.size(2) == head_dim
-    k = static_k
+    k = static_k.view(bsz, num_heads, -1, head_dim)
```
albertz, line 664: According to the doc: `static_k: (N*num_heads, S, E/num_heads)`.

Did you actually test this? This looks wrong. Shouldn't it be:
```
k = static_k.view(num_heads, bsz, -1, head_dim).transpose(0, 1)
```
?

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -651,60 +651,60 @@ def multi_head_attention_forward(
...
 
   if static_v is not None:
     assert static_v.size(0) == bsz * num_heads
     assert static_v.size(2) == head_dim
-    v = static_v
+    v = static_v.view(bsz, num_heads, -1, head_dim)
```
albertz, line 669: Same here.

pytorch_to_returnn/torch/nn/functional.py
albertz: Ah ok. Then just leave it for now. (The RETURNN generated code for this is anyway a no-op.)

pytorch_to_returnn/torch/nn/functional.py
albertz: Hm maybe not, and this is fine. I don't really understand from the documentation.

pytorch_to_returnn/torch/nn/functional.py
vieting: I added comments accordingly

albertz (approved):

pytorch_to_returnn/torch/nn/functional.py
vieting: No I didn't test this. I'm also not sure but I thought it could be done this way as they have `N` first in the doc

PR #41, opened 2021-04-30
assert correct torch shape of tensor entries

vieting:
Currently, it is not asserted that the `torch_shape` from `Module._get_output_shape_from_returnn` matches the actual torch shape. This is added here. This certainly will let existing tests using `_ConvNd` and `_ConvTransposedNd` fail. I'll add a fix in a separate PR.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/module.py
+++ b/pytorch_to_returnn/torch/nn/modules/module.py
@@ -636,6 +636,9 @@ class Module:
       returnn_output_np = returnn_output_np_.transpose(*[
         returnn_output_tensor_entry.returnn_axis_from_torch_axis[i] for i in range(returnn_output_np_.ndim)])
       torch_out_np = torch_mod_call.orig_outputs_flat[out_idx].detach().cpu().numpy()
+      assert torch_out_np.shape == tuple(int(i) for i in returnn_output_tensor_entry.tensor().shape), (
+        f"torch shape of TensorEntry does not match real torch shape: "
+        f"{torch_out_np.shape} vs. {returnn_output_tensor_entry}")
       error_msg_info = [f"RETURNN layer: {call.returnn_layer}", f"Torch module: {torch_mod}"]
       if returnn_output_np.shape != torch_out_np.shape:
         error_msg_info += [


albertz:
> This certainly will let existing tests using `_ConvNd` and `_ConvTransposedNd` fail.

Why? Only if they use striding.

> I'll add a fix in a separate PR.

If this causes them to fail, this should be part of this PR, not a separate PR.

pytorch_to_returnn/torch/nn/modules/module.py
```diff
@@ -636,6 +636,9 @@ def check_call_returnn_outputs_to_prev_torch(call: CallEntry, *, update_ops: Lis
       returnn_output_np = returnn_output_np_.transpose(*[
         returnn_output_tensor_entry.returnn_axis_from_torch_axis[i] for i in range(returnn_output_np_.ndim)])
       torch_out_np = torch_mod_call.orig_outputs_flat[out_idx].detach().cpu().numpy()
+      assert torch_out_np.shape == tuple(int(i) for i in returnn_output_tensor_entry.tensor().shape), (
+        f"torch shape of TensorEntry does not match real torch shape: "
+        f"{torch_out_np.shape} vs. {returnn_output_tensor_entry}")
```
albertz, line 641: The message should also include `returnn_output_tensor_entry.tensor().shape`.

vieting: It's part of what is printed now, but we can also reduce it to only `returnn_output_tensor_entry.tensor().shape` if you prefer.

vieting:
> Why? Only if they use striding.
Yes, that's true.
>If this causes them to fail, this should be part of this PR, not a separate PR.
Ok, that's also fine.

albertz:
No, both together are good.


vieting ***@***.***> schrieb am Fr., 30. Apr. 2021, 15:49:

pytorch_to_returnn/torch/nn/modules/module.py
albertz: No, both together are good.

pytorch_to_returnn/torch/nn/modules/conv.py
```diff
@@ -190,6 +207,23 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
...
+    The basic returnn_axis_from_torch_axis should be correct, however, if the size of a dynamic axis changes (e.g. due
+    to strides and/or padding), this is not covered in the base method and we fix it here.
+    """
+    torch_shape, returnn_axis_from_torch_axis = super(_ConvNd, self)._get_output_shape_from_returnn(
+      inputs_flat=inputs_flat, layer=layer)
+    torch_axis_from_returnn_axis = {j: i for i, j in returnn_axis_from_torch_axis.items()}
+    torch_shape = list(torch_shape)
+    assert len(layer.output.get_dynamic_axes()) == self.nd
+    for idx, ax in enumerate(layer.output.get_dynamic_axes()):
+      torch_ax = torch_axis_from_returnn_axis[ax]
```
albertz, line 222: Same as above.
```suggestion
    torch_shape = list(torch_shape)
    for idx in range(self.nd):
      torch_ax = idx + 1
```

pytorch_to_returnn/torch/nn/modules/conv.py
```diff
@@ -116,6 +116,23 @@ def import_params_torch_to_returnn(self, *, layer: LayerBase, torch_module: _Con
...
```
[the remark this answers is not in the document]
vieting: Yes, the assertion has to be removed.

The simplification works (with `torch_ax = idx + 1`) when only dynamic axes are affected, but for other spatial axes the `torch_shape` is already correct from the base method. I now compute the `torch_shape` completely here as it is straight-forward.

pytorch_to_returnn/torch/nn/modules/module.py
albertz: Where is `returnn_output_tensor_entry.tensor().shape`?

vieting: In `TensorEntry.__repr__`

albertz (approved):

PR #42, opened 2021-05-05
fix torch shape for multiple modules

vieting:
In #41 we introduced asserting the correct torch shape of tensor entries. There are multiple modules which have issues here. This PR fixes this by adding `_get_output_shape_from_returnn` for a number of modules.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -294,6 +294,27 @@ class Slice(Module):
       "slice_start": self.start, "slice_end": self.stop, "slice_step": self.step,
       "from": self._get_input_layer_name(input)}
 
+  def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    """
+    The size of the dynamic axes might be changed, so we have to take care of this here for the torch shape.
+    """
+    torch_shape, returnn_axis_from_torch_axis = super(Slice, self)._get_output_shape_from_returnn(
+      inputs_flat=inputs_flat, layer=layer)
+    assert len(inputs_flat) == 1
+    torch_shape = list(inputs_flat[0].shape)
+    start = self.start or 0
+    stop = self.stop or torch_shape[self.axis]
+    step = self.step or 1
+    if start < 0:
+      start += torch_shape[self.axis]
+    if stop < 0:
+      stop += torch_shape[self.axis]
+    assert 0 <= start <= torch_shape[self.axis] - 1
+    assert 0 <= stop <= torch_shape[self.axis]
+    torch_shape[self.axis] = len(range(start, stop, step))
+    return tuple(torch_shape), returnn_axis_from_torch_axis
+
 
 class Stack(Module):
   """
--- a/pytorch_to_returnn/torch/nn/modules/padding.py
+++ b/pytorch_to_returnn/torch/nn/modules/padding.py
@@ -1,8 +1,8 @@
 
 from ...tensor import Tensor
 from .module import Module
-from typing import Union, Tuple, Optional, List, Dict
 from returnn.tf.layers.basic import LayerBase
+from typing import Union, Tuple, Optional, List, Dict
 from .utils import _pair, _quadruple, _ntuple
 from ..common_types import _size_2_t, _size_4_t, _size_6_t
 from ....naming import Naming
@@ -65,8 +65,7 @@ class GenericPadNd(Module):
   def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
                                      ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
     """
-    The basic returnn_axis_from_torch_axis should be correct, however, the torch shape is not adapted in the base method
-    and we fix it here.
+    The size of the dynamic axes might be changed, so we have to take care of this here for the torch shape.
     """
     torch_shape, returnn_axis_from_torch_axis = super(GenericPadNd, self)._get_output_shape_from_returnn(
       inputs_flat=inputs_flat, layer=layer)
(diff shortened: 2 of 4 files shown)


pytorch_to_returnn/torch/nn/modules/activation.py
```diff
@@ -11,6 +12,17 @@ class _ActivationReturnn(Module):
   def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
     return {"class": "activation", "activation": self.func_name, "from": self._get_input_layer_name(input)}
 
+  def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    """
+    The basic returnn_axis_from_torch_axis should be correct, however, the torch shape is sometimes incorrectly adapted
+    in the base method and we fix it here.
```
albertz, line 19: I don't exactly understand. In what cases? And why?

And this sounds like a bug to me. Is it not? If this is a bug in the base method, we should fix it directly there?

Also, what do you mean "fix it here"? You don't actually do anything here?

Btw, I think there is one case where it might actually change the shape (order of dims), and that is if you use "softmax". Not sure. It might move the feature dim to the end.

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -60,6 +60,19 @@ def import_params_torch_to_returnn(self, *, layer: LinearLayer, torch_module: Li
     if self.bias is not None:
       layer.params["b"].load(torch_module.bias.detach().numpy(), session=session)
 
+  def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    """
+    The basic returnn_axis_from_torch_axis should be correct, however, the torch shape is not always correct in the base
+    method and we fix it here.
```
albertz, line 67: When is it not correct? And why?

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -78,6 +78,27 @@ def create_returnn_layer_dict(self, *inputs: Tensor):
       "class": "combine", "kind": self.kind,
       "from": [self._get_input_layer_name(input) for input in inputs]}
 
+  def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    """
+    The basic returnn_axis_from_torch_axis should be correct, however, the torch shape is not adapted in the base method
+    and we fix it here.
```
albertz, line 85: Why is the base code not correct here? I would assume it is.

albertz:
In all cases, you write:

> The basic returnn_axis_from_torch_axis should be correct, however, the torch shape is sometimes incorrectly adapted
> in the base method and we fix it here.

In many (almost all) of the cases, I don't see why that is.

And is this not a bug maybe of the base method, which we should simply fix?

vieting:
Yes, there was a bug in the base method. This is fixed now. Only padding and slicing still need their own `_get_output_shape_from_returnn` because they might change the dynamic dims.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -294,6 +294,27 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
...
 
+  def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    """
+    The size of the dynamic axes might be changed, so we have to take care of this here for the torch shape.
+    """
+    torch_shape, returnn_axis_from_torch_axis = super(Slice, self)._get_output_shape_from_returnn(
+      inputs_flat=inputs_flat, layer=layer)
+    assert len(inputs_flat) == 1
+    torch_shape = list(inputs_flat[0].shape)
```
albertz, line 305: If you make the hard assumption here that the order of the axis will not change by this layer, then you can simply also take `returnn_axis_from_torch_axis` from the inputs, and not calculate it. Or rather, you should do this, for consistency, and also simplicity.

pytorch_to_returnn/torch/nn/modules/padding.py
```diff
@@ -36,6 +37,19 @@ def create_returnn_layer_dict(self, input: Tensor):
...
 
+  def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    """
+    The size of the dynamic axes might be changed, so we have to take care of this here for the torch shape.
+    """
+    torch_shape, returnn_axis_from_torch_axis = super(GenericPadNd, self)._get_output_shape_from_returnn(
+      inputs_flat=inputs_flat, layer=layer)
+    assert len(inputs_flat) == 1
+    torch_shape = list(inputs_flat[0].shape)
```
albertz, line 48: Same here for `returnn_axis_from_torch_axis`.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: Yes, done.

vieting:
To me it seems that all comments were adressed. Could you please check @albertz?

vieting:
There are now multiple test cases failing at
```
 File "/home/runner/work/pytorch-to-returnn/pytorch-to-returnn/pytorch_to_returnn/converter/converter.py", line 158, in Converter._run_traced_orig_torch
    line: assert isinstance(out_wrapped, WrappedTorchTensor)
```
I cannot reproduce this locally with the same RETURNN version (current master), python 3.8.0 (here 3.8.10), tf 2.3.0 (here 2.3.1) and same numpy. Do you have an idea?

albertz:
Maybe new PyTorch (1.9) version? (Not much time currently to investigate...)

albertz:
Yep, this problems is related to PyTorch 1.9. For our CI tests, I now explicitly use PyTorch 1.8 (via #45). So please rebase, then the tests should pass again (unless your PR introduces some other bugs).

I added the test for PyTorch 1.9 in #46 but this needs some fixing (independently from this PR).

albertz (approved):

PR #43, opened 2021-05-05
allow setting shape for returnn dummy input

vieting:
The returnn dummy input uses arbitrary sizes for the dynamic axes. I had a case where the size of the time dim was reduced by convolutions such that the chosen arbitrary size was too small (see added test case). For such situations, I added the option `returnn_dummy_input_shape` for the `Converter` so that the dummy input shape can be manually set.


vieting:
#49 fixes the test that was added and this PR is now ready for review.

albertz (approved):

PR #52, opened 2021-08-18
allow dict/list inputs

vieting:
This PR makes it possible to use dicts or lists of tensors as input to the PyTorch models that should be converted. It was already possible to have them as input/output of a forward function (see the new `test_dict_in_forward` which already passes without the suggested modifications). Now, it is also possible to have them as the initial input to the model.

### DIFF
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -7,7 +7,7 @@ import numpy
 import types
 import tempfile
 from pytorch_to_returnn.pprint import pprint
-from typing import Callable, Optional, Dict, Any, Tuple
+from typing import Callable, Optional, Dict, Any, Tuple, List, Union
 from returnn.tf.util.data import Data
 from pytorch_to_returnn import torch as torch_returnn
 from pytorch_to_returnn.import_wrapper import wrapped_import_torch_traced, wrapped_import_torch_returnn
@@ -16,6 +16,7 @@ from pytorch_to_returnn.naming import Naming
 
 
 ModelFuncType = Callable[[Optional[Callable[[str], types.ModuleType]], torch.Tensor], torch.Tensor]
+InputType = Union[numpy.ndarray, Dict[Any, numpy.ndarray], List[numpy.ndarray]]
 
 
 @contextlib.contextmanager
@@ -62,7 +63,7 @@ class Converter:
 
   def __init__(self,
                model_func: ModelFuncType, *,
-               inputs: numpy.ndarray,
+               inputs: InputType,
                inputs_data_kwargs: Optional[Dict[str, Any]] = None,
                returnn_dummy_input_shape: Optional[Tuple[int]] = None,
                use_non_wrapped_reference: bool = True,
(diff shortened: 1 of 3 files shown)


pytorch_to_returnn/converter/converter.py
```diff
@@ -81,17 +82,37 @@ def __init__(self,
...
-      assert len(inputs.shape) >= 2  # (batch,feature|channel,...)
-      if inputs_data_kwargs.get("batch_dim_axis", 0) == 0:
-        inputs_data_kwargs["feature_dim_axis"] = 1  # Torch uses batch-feature-major by default
-    if "shape" not in inputs_data_kwargs:
-      # Assume feature is static, all other dynamic.
-      # Assume batch-major.
-      inputs_data_kwargs["shape"] = [
-        inputs.shape[i] if i == inputs_data_kwargs["feature_dim_axis"] else None
-        for i in range(1, len(inputs.shape))]
+    assert type(self._inputs_np) in (numpy.ndarray, dict, list), (
```
albertz, line 85: This should not be restricted. You should just use the `nest` module for this (see other code which uses `nest`).

pytorch_to_returnn/converter/converter.py
```diff
@@ -81,17 +82,37 @@ def __init__(self,
...
+      "Type not supported as input: {}".format(type(inputs_data_kwargs)))
+    # inputs_data_kwargs should always be of type Dict[str|int, Dict[str, Any]].
+    # If input is array, the only key is "data". If input is a list, the keys are 0, 1, ...
+    if inputs_data_kwargs is None:
+      if isinstance(self._inputs_np, numpy.ndarray):
+        inputs_data_kwargs = {"data": {}}
+      elif isinstance(self._inputs_np, dict):
+        inputs_data_kwargs = {k: {} for k in self._inputs_np}
+      elif isinstance(self._inputs_np, list):
+        inputs_data_kwargs = {k: {} for k in range(len(self._inputs_np))}
```
albertz, line 95: This is all too complicated and too specific. Just use the `nest` module.

pytorch_to_returnn/converter/converter.py
```diff
@@ -81,17 +82,37 @@ def __init__(self,
...
+    if isinstance(inputs_data_kwargs, list):
+      inputs_data_kwargs = {i: x for i, x in enumerate(inputs_data_kwargs)}
+    if isinstance(self._inputs_np, numpy.ndarray) and inputs_data_kwargs.keys() != {"data"}:
+      inputs_data_kwargs = {"data": inputs_data_kwargs}
+    assert isinstance(inputs_data_kwargs, dict)
```
albertz, line 100: This check is wrong. I assume you want to assert the same nested structure as `inputs`? Just use `nest`.

pytorch_to_returnn/converter/converter.py
```diff
@@ -122,15 +143,37 @@ def returnn_net_dict(self) -> Dict[str, Dict[str, Any]]:
...
 
-  def _make_tf_feed_dict(self, input: Data):
-    assert input.batch_ndim == len(self._inputs_np.shape)
-    assert all(input.batch_shape[i] in {None, self._inputs_np.shape[i]} for i in range(input.batch_ndim))
-    n_batch = self._inputs_np.shape[input.batch_dim_axis]
-    d = {input.placeholder: self._inputs_np}
-    for i, size in input.size_placeholder.items():
-      d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
+  def _make_tf_feed_dict(self, inputs: Dict[Any, Data]):
+    def _as_list(obj):
```
albertz, line 147: Do not implement such a function here. Just use `nest`.

pytorch_to_returnn/converter/converter.py
```diff
@@ -197,9 +240,21 @@ def _run_torch_returnn_drop_in(self):
             keep_orig_module_io_tensors=True,  # it's only symbolic anyway in TF
             import_params_from_torch_namespace=self._torch_namespace) as naming:
         assert isinstance(naming, Naming)
-        in_returnn = torch_returnn.from_numpy(self._inputs_np.copy())
-        assert isinstance(in_returnn, torch_returnn.Tensor)
-        x = naming.register_input(in_returnn, Data("data", **self._returnn_in_data_dict))
+        in_returnn = self._get_torch_inputs(torch_returnn)
+        if isinstance(in_returnn, torch_returnn.Tensor):
```
albertz, line 244: This is all again too complicated. Use `nest`.

pytorch_to_returnn/converter/converter.py
```diff
@@ -248,17 +303,19 @@ def _run_returnn_standalone_net_dict(self):
     with make_scope() as session:
       from returnn.config import Config
       from returnn.tf.network import TFNetwork
+      extern_data = {
+        str(i) if i else "data": dict({"available_for_inference": True}, **v)
+        for i, v in enumerate(self._returnn_in_data_dict.values())}
```
albertz, line 308: The order of `values` is non-deterministic in older Python versions. I'm not sure if this is good then.

pytorch_to_returnn/torch/nn/modules/module.py
```diff
@@ -933,27 +934,35 @@ def _make_returnn_dummy_input(cls, data: Data) -> numpy.ndarray:
         shape[j] = some_primes[i]
     return numpy.zeros(shape, dtype=data.dtype)
 
-  def _returnn_dummy_call(self, *returnn_inputs: Dict[str, Any]) -> Naming:
+  def _returnn_dummy_call(self, returnn_inputs: Dict[Any, Dict[str, Any]], original_inputs: InputType) -> Naming:
     from pytorch_to_returnn.torch import from_numpy
     naming = Naming.get_instance()
     returnn_datas = []
-    for i, kwargs in enumerate(returnn_inputs):
+    for i, kwargs in enumerate(returnn_inputs.values()):
```
albertz, line 941: Same issue with `values`.

pytorch_to_returnn/torch/nn/modules/module.py
```diff
@@ -933,27 +934,35 @@ def _make_returnn_dummy_input(cls, data: Data) -> numpy.ndarray:
...
       returnn_datas.append(x)
     dummy_inputs_np = [self._make_returnn_dummy_input(x) for x in returnn_datas]
     dummy_inputs_torch = [from_numpy(x) for x in dummy_inputs_np]
     for i in range(len(returnn_inputs)):
       naming.register_input(tensor=dummy_inputs_torch[i], returnn_data=returnn_datas[i])
-    out = self(*dummy_inputs_torch)
+    if isinstance(original_inputs, numpy.ndarray):
+      dummy_inputs_torch = dummy_inputs_torch[0]
+    if isinstance(original_inputs, dict):
+      dummy_inputs_torch = {k: v for k, v in zip(original_inputs.keys(), dummy_inputs_torch)}
```
albertz, line 954: Do not use such code. Use `nest` instead.

albertz (changes requested):
Many issues where you should use `nest` instead.

But more general: I don't understand, why does the `model_func` need this? The `model_func` is just intended as a simple wrapper to test the conversion. For nothing more. You don't really need this. Or why do you need this?

vieting:
>Many issues where you should use nest instead.

Thanks for the hint, I'll check that and update accordingly.

>But more general: I don't understand, why does the model_func need this? The model_func is just intended as a simple wrapper to test the conversion. For nothing more. You don't really need this. Or why do you need this?

E.g. the module in [here](https://github.com/fgnt/padertorch/blob/master/padertorch/contrib/examples/source_separation/pit/model.py) has a dict as input. In order to convert it, I create a model_func in which I load the module and forward the inputs to the model_function through the module. Therefore I need the dict as input. Or do you have an easier way?

albertz:
> E.g. the module in [here](https://github.com/fgnt/padertorch/blob/master/padertorch/contrib/examples/source_separation/pit/model.py) has a dict as input. In order to convert it, I create a model_func in which I load the module and forward the inputs to the model_function through the module. Therefore I need the dict as input. Or do you have an easier way?

The `model_func` is just for testing. It is not meant to wrap a module. It just needs the numpy array for the test, nothing else.

See [this example](https://github.com/rwth-i6/pytorch-to-returnn/blob/main/examples/parallel_wavegan/pytorch_to_returnn.py). It even loads a custom dict from a config file and passes it to the module. But the `model_func` function just gets the numpy array, nothing else.

vieting:
Yes, I think you're right. This is probably not necessary.

PR #55, opened 2021-10-07
Added test for using tensor.view for removing an axis

dierkes-j:
I ran into a problem using the method tensor.view in order to remove an axis with dimension 1. I think the output shape of the tensor is permuted in an undesired manner in reshape. I created a test to reproduce the problem.


albertz:
I fixed this now via #57, which is merged now, so this PR here is not needed anymore.

PR #66, opened 2021-10-20
add tensor.new_full()

vieting:


vieting:
After #88 was merged, I rebased and use `torch.nn.functional.full` which was introduced in the meantime.

albertz (approved):

PR #69, opened 2021-10-27
new zeros with dynamic axes

vieting:
`tensor.new_zeros()` currently only works for static dims. I added a test case to demonstrate this. I think the comparison of dim tags should be fixed as in the commit, but still the axes are not matched correctly and RETURNN inserts an additional extra axis.

Small extract of the error:
```
AssertionError: 
Not equal to tolerance rtol=0, atol=0.0005

(shapes (3, 5, 11), (5, 3, 5, 11) mismatch)
```


albertz:
Can you rebase this and check if it works already? I think it should work already.

albertz:
Continued work in #88.

PR #71, opened 2021-11-12
Add support for PackedSequences

vieting:
For an example from padertorch that I'd like to import, we need support for `PackedSequence`. I implemented a first version of what we discussed and would appreciate your feedback on it, @albertz :)

### DIFF
--- a/pytorch_to_returnn/import_wrapper/wrap.py
+++ b/pytorch_to_returnn/import_wrapper/wrap.py
@@ -38,6 +38,11 @@ def wrap(obj, *, name: str, ctx: WrapCtx):
       pass  # keep as-is
     elif type(obj) != type:  # explicitly do not check sub-types, e.g. pybind11_type
       pass
+    elif issubclass(obj, tuple):  # e.g. namedtuple
+      # Don't wrap this object. This was covered already in _nested_transform().
+      # We might run into issues with e.g. namedtuple otherwise because the __bases__ cannot always be added to the
+      # WrappedClass and we just ignore that, then subsequently isinstance(wrapped_object, tuple) would return False
+      pass
     elif obj.__module__ and ctx.should_wrap_mod(obj.__module__):
       # If this is an indirect module, but the type is part of a submodule which we want to transform,
       # explicitly check the import.
@@ -51,6 +56,11 @@ def wrap(obj, *, name: str, ctx: WrapCtx):
       obj = make_wrapped_object(obj, name=name, ctx=ctx)
     elif not getattr(type(obj), "__module__", None) or not ctx.should_wrap_mod(type(obj).__module__):
       pass  # Don't wrap this object.
+    elif issubclass(type(obj), tuple):  # e.g. namedtuple
+      # Don't wrap this object. This was covered already in _nested_transform().
+      # We might run into issues with e.g. namedtuple otherwise because the __bases__ cannot always be added to the
+      # WrappedClass and we just ignore that, then subsequently isinstance(wrapped_object, tuple) would return False
+      pass
     elif isinstance(obj, ctx.keep_as_is_types):  # blacklist
       pass  # keep as-is
     else:
@@ -75,9 +85,12 @@ def unwrap(obj):
 
 
 def _nested_transform(obj, transform):
+  if isinstance(obj, WrappedObject):
+    return obj
   if type(obj) == tuple:
     return tuple([transform(item) for item in obj])
-  assert not isinstance(obj, tuple)  # namedtuple or so not implemented yet...
+  if issubclass(type(obj), tuple) and type(obj) is not tuple:
+    return type(obj)(*(transform(item) for item in obj))
   if type(obj) == list:
     return [transform(item) for item in obj]
   assert not isinstance(obj, list)  # custom subclasses of list not implemented yet...
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -125,6 +125,7 @@ class SizeValue(int):
   We extend this, to store extra information, e.g. such that this reflects the batch-dim.
   """
   is_batch_dim: bool = False
+  merged_dims: list[SizeValue] = []
 
   def __repr__(self):
     res = super(SizeValue, self).__repr__()
--- a/pytorch_to_returnn/torch/nn/modules/module.py
+++ b/pytorch_to_returnn/torch/nn/modules/module.py
@@ -54,6 +54,7 @@ class Module:
   # All derived classes here, which exist in PyTorch as well (e.g. torch.nn.Conv1d etc),
   # should have this set to True.
   is_original_torch_module: bool = True
+  _forward_feed_dict_deps: bool = False  # for internal usage
 
   _wrapped_class_cache = {}  # cls -> WrappedClass
 
@@ -459,8 +460,8 @@ class Module:
       def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]:
         return self_.create_returnn_layer_dict(*inputs, **kwargs)
 
-      def make_structured_returnn_output(self, output):
-        return self_.make_structured_returnn_output(output)
+      def make_structured_returnn_output(self, output, *inputs_args, **inputs_kwargs):
+        return self_.make_structured_returnn_output(output, *inputs_args, **inputs_kwargs)
 
       def get_returnn_name(self) -> str:
         return self_.get_returnn_name()
@@ -473,7 +474,8 @@ class Module:
   def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]:
     raise Exception("should not get here")
 
-  def make_structured_returnn_output(self, output: Tensor) -> Union[Tensor, Tuple[Tensor], Any]:
+  def make_structured_returnn_output(self, output: Tensor, *inputs: Tensor, **inputs_kwargs
+                                     ) -> Union[Tensor, Tuple[Tensor], Any]:
     """
     This can be overridden alongside with `create_returnn_layer_dict`.
     In case the original `forward` would return not a single tensor but some tuple or dict or other nested structure,
@@ -578,7 +580,8 @@ class Module:
     out_size_feed_dict.update(sizes_feed_dict)
     tensor.validated_to_torch_tf_feed_dict.update(out_size_feed_dict)
     tensor.validated_to_torch_tf_sizes_feed_dict = out_size_feed_dict
-
+    if any(call_.module.module._forward_feed_dict_deps for call_ in tensor.output_from_calls):
+      tensor.validated_to_torch_tf_feed_dict.update(feed_dict)
     # Check now against Torch reference.
     torch_axis_from_returnn_axis = {i: j for (j, i) in tensor.returnn_axis_from_torch_axis.items()}
     assert len(torch_values.shape) == tensor.returnn_data.batch_ndim
@@ -623,15 +626,20 @@ class Module:
       idx_repr = f" {i + 1}/{len(call.outputs_flat)}" if len(call.outputs_flat) > 1 else ""
       print(f"**** validate: add call {call} output{idx_repr} tensor {x}")
     out, _ = session.run((
-      [(x.returnn_data.placeholder, x.returnn_data.size_placeholder.as_dict()) for x in call.outputs_flat], update_ops),
+      [(x.returnn_data.placeholder, x.returnn_data.size_placeholder.as_dict()) if x is not None else () for
+       x in call.outputs_flat], update_ops),
       feed_dict=feed_dict)
     for out_idx, returnn_output_tensor_entry in enumerate(call.outputs_flat):
+      if returnn_output_tensor_entry is None:
+        continue
       idx_repr = f" {out_idx + 1}/{len(call.outputs_flat)}" if len(call.outputs_flat) > 1 else ""
       returnn_output_np_, output_sizes = out[out_idx]
       assert isinstance(returnn_output_np_, numpy.ndarray)
       returnn_output_tensor_entry.validated_to_torch = True
       returnn_output_tensor_entry.validated_to_torch_tf_feed_dict = {
         returnn_output_tensor_entry.returnn_data.placeholder: returnn_output_np_}
+      if call.module.module._forward_feed_dict_deps:
+        returnn_output_tensor_entry.validated_to_torch_tf_feed_dict.update(feed_dict)
       out_size_feed_dict = {
         returnn_output_tensor_entry.returnn_data.size_placeholder[i]: output_sizes[i]
         for i in returnn_output_tensor_entry.returnn_data.size_placeholder}
--- a/pytorch_to_returnn/torch/nn/modules/rnn.py
+++ b/pytorch_to_returnn/torch/nn/modules/rnn.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import math
 import warnings
 import numbers
+from collections import namedtuple
 from typing import Dict, Any, Optional, List, Tuple, Union
 import tensorflow as tf
 from returnn.tf.layers.basic import LayerBase, SubnetworkLayer
@@ -11,10 +12,33 @@ from .module import Module
 from ..parameter import Parameter
 from ...tensor import Tensor
 from .. import init
+from ....naming import Naming, TensorEntry
 
 
-class PackedSequence:  # dummy -- not yet implemented...
-  pass
+PackedSequence_ = namedtuple(
+  'PackedSequence',
+  ['data', 'batch_sizes', 'sorted_indices', 'unsorted_indices'],
+  defaults=[None] * 4)
+
+
+class PackedSequence(PackedSequence_):
+  """
+  Wrapper for torch's PackedSequence
+
+  We connect each PackedSequence with a batched RETURNN tensor. This currently contains some assumptions which might not
+  always be satisfied.
+  """
+
+  def get_batched_tensor(self) -> Tensor:
+    from .shape import FlattenBatch
+    naming = Naming.get_instance()
+    tensor_entry = naming.tensors[self.data]
+    assert isinstance(tensor_entry, TensorEntry)
+    for call_entry in tensor_entry.output_from_calls:
+      if isinstance(call_entry.module.module, FlattenBatch):
+        return call_entry.orig_inputs_args[0]
+    from .shape import UnflattenBatch
+    return UnflattenBatch()(self.data)
 
 
 def apply_permutation(tensor: Tensor, permutation: Tensor, dim: int = 1) -> Tensor:
--- a/pytorch_to_returnn/torch/nn/modules/rnn.py
+++ b/pytorch_to_returnn/torch/nn/modules/rnn.py
@@ -205,6 +205,8 @@ class LSTM(RNNBase):
     return apply_permutation(hx[0], permutation), apply_permutation(hx[1], permutation)
 
   def create_returnn_layer_dict(self, input: Tensor, hx: Optional[Tensor] = None) -> Dict[str, Any]:
+    if isinstance(input, PackedSequence):
+      input = input.get_batched_tensor()
     if not self.bidirectional and self.num_layers == 1:
       d = {
         "class": "rec", "unit": "nativelstm2", "from": self._get_input_layer_name(input),
@@ -247,7 +249,19 @@ class LSTM(RNNBase):
         "class": "subnetwork", "from": input_layer_name, "subnetwork": subnet_dict}
     return d
 
-  def make_structured_returnn_output(self, output: Tensor) -> Union[Tensor, Tuple[Tensor], Any]:
+  def _get_output_shape_from_returnn(self,
+                                     inputs_flat: List[Tensor], layer: LayerBase
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    naming = Naming.get_instance()
+    call_entry = naming.module_call_stack[-1]
+    assert call_entry.module.module is self
+    orig_input = call_entry.orig_inputs_args[0]
+    if isinstance(orig_input, PackedSequence):
+      orig_input = orig_input.get_batched_tensor()
+    return self._base_get_output_shape_from_returnn(inputs_flat=[orig_input], layer=layer)
+
+  def make_structured_returnn_output(self, output: Tensor, input: Tensor, hx: Optional[Tensor] = None
+                                     ) -> Union[Tensor, Tuple[Tensor], Any]:
     hs, cs = [], []
     dim = output.shape[-1]
     if not self.bidirectional and self.num_layers == 1:
@@ -266,6 +280,9 @@ class LSTM(RNNBase):
     from .operator import Stack
     h = Stack(dim=0)(*hs)
     c = Stack(dim=0)(*cs)
+    if isinstance(input, PackedSequence):
+      from ..utils.rnn import pack_padded_sequence_with_batch_sizes
+      output = pack_padded_sequence_with_batch_sizes(output, input.batch_sizes)
     return output, (h, c)
 
   def check_returnn_layer(self, layer: LayerBase):
--- a/pytorch_to_returnn/torch/nn/modules/shape.py
+++ b/pytorch_to_returnn/torch/nn/modules/shape.py
@@ -315,7 +315,7 @@ class Split(Module):
       d["size_splits"] = self.size_splits
     return d
 
-  def make_structured_returnn_output(self, output: Tensor) -> List[Tensor]:
+  def make_structured_returnn_output(self, output: Tensor, input: Tensor) -> List[Tensor]:
     from .operator import GetSublayer
     if self.num_splits is not None:
       num_splits = self.num_splits
--- a/pytorch_to_returnn/torch/nn/modules/shape.py
+++ b/pytorch_to_returnn/torch/nn/modules/shape.py
@@ -15,6 +15,7 @@ import operator
 from returnn.tf.layers.basic import LayerBase, MergeDimsLayer, FlattenBatchLayer
 from .module import Module
 from ...tensor import Tensor, Size
+from ..._C import SizeValue
 from ....naming import Naming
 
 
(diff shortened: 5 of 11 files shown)


pytorch_to_returnn/import_wrapper/wrap.py
```diff
@@ -51,6 +53,8 @@ def wrap(obj, *, name: str, ctx: WrapCtx):
       obj = make_wrapped_object(obj, name=name, ctx=ctx)
     elif not getattr(type(obj), "__module__", None) or not ctx.should_wrap_mod(type(obj).__module__):
       pass  # Don't wrap this object.
+    elif issubclass(type(obj), tuple):  # e.g. namedtuple
+      pass  # Don't wrap this object.
```
albertz, line 57: I don't exactly understand this. Why is it a problem? This seems wrong.

pytorch_to_returnn/torch/_C.py
```diff
@@ -125,6 +125,7 @@ class SizeValue(int):
   We extend this, to store extra information, e.g. such that this reflects the batch-dim.
   """
   is_batch_dim: bool = False
+  merged_dims = []
```
albertz, line 128: What is the type?

pytorch_to_returnn/torch/nn/modules/shape.py
```diff
@@ -323,6 +324,57 @@ def make_structured_returnn_output(self, output: Tensor) -> List[Tensor]:
...
+                                     inputs_flat: List[Tensor], layer: Union[MergeDimsLayer, FlattenBatchLayer]
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    naming = Naming.get_instance()
+    assert len(inputs_flat) == 1
+    x = inputs_flat[0]
+    x_entry = naming.register_tensor(x)
+    dims_to_merge = [
+      dim for i, dim in enumerate(x.shape) if i in x_entry.returnn_data.get_axes_from_description(["B" + self.axis])]
+    assert len(dims_to_merge) == 2
+    merged_dim = SizeValue(dims_to_merge[0] * dims_to_merge[1])
```
albertz, line 350: This is not exactly correct. It would be the sum of all seq lengths.

pytorch_to_returnn/torch/nn/utils/__init__.py
```diff
@@ -1,2 +1,3 @@
 
+from .rnn import *
```
albertz, line 2: This looks wrong. Why do you need this?
Maybe you just want this instead?
```suggestion
from . import rnn
```

tests/test_layers.py
```diff
@@ -341,6 +341,94 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
+  """
+  n_batch, n_time, n_feat = 3, 5, 7
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    if typing.TYPE_CHECKING or not wrapped_import:
+      import torch
+    else:
+      torch = wrapped_import("torch")
+
+    h = torch.nn.utils.rnn.pack_padded_sequence(inputs, [n_time] * n_batch)
```
albertz, line 356: I think it would be more interesting to use seq lengths like `[5, 4, 3]` or so.

tests/test_layers.py
```diff
@@ -341,6 +341,94 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
```
albertz, line 422: Here also, use seq length `[5, 4, 3]` or so.

tests/test_layers.py
```diff
@@ -341,6 +341,94 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
```
albertz, line 401: Seq lengths `[5, 4, 3]`.

pytorch_to_returnn/import_wrapper/wrap.py
vieting: A `PackedSequence` should not be wrapped as a `WrappedObject`, and I excluded `namedtuples` here in general.

albertz: But I don't understand this. Why should it not be wrapped? And why not `namedtuple`s in general?

vieting: I think we discussed this to allow unpacking of `PackedSequence`s. I'm not sure about this, though. However, `make_wrapped_object` will fail for a `PackedSequence` in `_run_traced_orig_torch`:

```
  File "../pytorch-to-returnn/tests/test_layers.py", line 1302, in <module>
    line: globals()[arg]()  # assume function and execute
    locals:
      globals = <builtin> <built-in function globals>
      arg = <local> 'test_packed_sequence_1', len = 22
  File "../pytorch-to-returnn/tests/test_layers.py", line 362, in test_packed_sequence_1
    line: verify_torch_and_convert_to_returnn(model_func, inputs=x, inputs_data_kwargs={
              "shape": (None, n_feat), "time_dim_axis": 0, "batch_dim_axis": 1, "feature_dim_axis": 2})
    locals:
      verify_torch_and_convert_to_returnn = <global> <function verify_torch_and_convert_to_returnn at 0x7f933798a4c0>
      model_func = <local> <function test_packed_sequence_1.<locals>.model_func at 0x7f933798d9d0>
      inputs = <not found>
      x = <local> array([[[ 0.49671414, -0.1382643 ,  0.64768857,  1.5230298 ,
                           -0.23415338, -0.23413695,  1.5792128 ],
                          [ 0.7674347 , -0.46947438,  0.54256004, -0.46341768,
                           -0.46572974,  0.24196227, -1.9132802 ],
                          [-1.7249179 , -0.5622875 , -1.0128311 ,  0.31424734,
                           -0.9080..., len = 5, _[0]: {len = 3, _[0]: {len = 7}}
      inputs_data_kwargs = <not found>
      n_feat = <local> 7
  File "/u/vieting/testing/20210714_dfg_kickoff/pytorch-to-returnn/pytorch_to_returnn/converter/converter.py", line 328, in verify_torch_and_convert_to_returnn
    line: converter.run()
    locals:
      converter = <local> <pytorch_to_returnn.converter.converter.Converter object at 0x7f93cc825400>
      converter.run = <local> <bound method Converter.run of <pytorch_to_returnn.converter.converter.Converter object at 0x7f93cc825400>>
  File "/u/vieting/testing/20210714_dfg_kickoff/pytorch-to-returnn/pytorch_to_returnn/converter/converter.py", line 114, in Converter.run
    line: self._run_traced_orig_torch()
    locals:
      self = <local> <pytorch_to_returnn.converter.converter.Converter object at 0x7f93cc825400>
      self._run_traced_orig_torch = <local> <bound method Converter._run_traced_orig_torch of <pytorch_to_returnn.converter.converter.Converter object at 0x7f93cc825400>>
  File "/u/vieting/testing/20210714_dfg_kickoff/pytorch-to-returnn/pytorch_to_returnn/converter/converter.py", line 170, in Converter._run_traced_orig_torch
    line: out_wrapped = self._model_func(wrapped_import_torch_traced, wrapped_torch.from_numpy(self._inputs_np.copy()))
    locals:
      out_wrapped = <not found>
      self = <local> <pytorch_to_returnn.converter.converter.Converter object at 0x7f93cc825400>
      self._model_func = <local> <function test_packed_sequence_1.<locals>.model_func at 0x7f933798d9d0>
      wrapped_import_torch_traced = <global> <function wrapped_import_torch_traced at 0x7f933798a040>
      wrapped_torch = <local> <WrappedIndirectModule pytorch_to_returnn.import_wrapper._torch_traced.torch>
      wrapped_torch.from_numpy = <local> <class 'pytorch_to_returnn.import_wrapper.base_wrappers.function._VariableFunctionsClass.from_numpy'>
      self._inputs_np = <local> array([[[ 0.49671414, -0.1382643 ,  0.64768857,  1.5230298 ,
                                         -0.23415338, -0.23413695,  1.5792128 ],
                                        [ 0.7674347 , -0.46947438,  0.54256004, -0.46341768,
                                         -0.46572974,  0.24196227, -1.9132802 ],
                                        [-1.7249179 , -0.5622875 , -1.0128311 ,  0.31424734,
                                         -0.9080..., len = 5, _[0]: {len = 3, _[0]: {len = 7}}
      self._inputs_np.copy = <local> <built-in method copy of numpy.ndarray object at 0x7f93cc74da30>
  File "../pytorch-to-returnn/tests/test_layers.py", line 356, in test_packed_sequence_1.<locals>.model_func
    line: h = torch.nn.utils.rnn.pack_padded_sequence(inputs, [n_time] * n_batch)
    locals:
      h = <not found>
      torch = <local> <WrappedIndirectModule pytorch_to_returnn.import_wrapper._torch_traced.torch>
      torch.nn = <local> <WrappedIndirectModule pytorch_to_returnn.import_wrapper._torch_traced.torch.nn>
      torch.nn.utils = <local> <WrappedIndirectModule pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.utils>
      torch.nn.utils.rnn = <local> <WrappedIndirectModule pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.utils.rnn>
      torch.nn.utils.rnn.pack_padded_sequence = <local> <function pack_padded_sequence at 0x7f93379a00d0>
      inputs = <local> tensor([[[ 0.4967, -0.1383,  0.6477,  1.5230, -0.2342, -0.2341,  1.5792],
                                [ 0.7674, -0.4695,  0.5426, -0.4634, -0.4657,  0.2420, -1.9133],
                                [-1.7249, -0.5623, -1.0128,  0.3142, -0.9080, -1.4123,  1.4656]],

                               [[-0.2258,  0.0675, -1.4247, -0.5444,  0.1109, -1.1510,  0.3757],..., len = 5, _[0]: {len = 3, _[0]: {len = 7}}
      n_time = <local> 5
      n_batch = <local> 3
  File "/u/vieting/testing/20210714_dfg_kickoff/pytorch-to-returnn/pytorch_to_returnn/import_wrapper/base_wrappers/function.py", line 16, in _VariableFunctionsClass.from_numpy
    line: res = wrap(res, name="%s(...)" % name, ctx=ctx)
    locals:
      res = <local> PackedSequence(data=tensor([[ 0.4967, -0.1383,  0.6477,  1.5230, -0.2342, -0.2341,  1.5792],
                            [ 0.7674, -0.4695,  0.5426, -0.4634, -0.4657,  0.2420, -1.9133],
                            [-1.7249, -0.5623, -1.0128,  0.3142, -0.9080, -1.4123,  1.4656],
                            [-0.2258,  0.0675, -1.4247, -0.5444,  0.1109, -1.1..., len = 4, _[0]: {len = 15, _[0]: {len = 7}}
      wrap = <local> <function wrap at 0x7f933798dd30>
      name = <local> 'pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.utils.rnn.pack_padded_sequence', len = 87
      ctx = <local> <WrapCtx 'pytorch_to_returnn.import_wrapper._torch_traced'>
  File "/u/vieting/testing/20210714_dfg_kickoff/pytorch-to-returnn/pytorch_to_returnn/import_wrapper/wrap.py", line 61, in wrap
    line: obj = make_wrapped_object(obj, name=name, ctx=ctx)
    locals:
      obj = <local> PackedSequence(data=tensor([[ 0.4967, -0.1383,  0.6477,  1.5230, -0.2342, -0.2341,  1.5792],
                            [ 0.7674, -0.4695,  0.5426, -0.4634, -0.4657,  0.2420, -1.9133],
                            [-1.7249, -0.5623, -1.0128,  0.3142, -0.9080, -1.4123,  1.4656],
                            [-0.2258,  0.0675, -1.4247, -0.5444,  0.1109, -1.1..., len = 4, _[0]: {len = 15, _[0]: {len = 7}}
      make_wrapped_object = <global> <function make_wrapped_object at 0x7f93379e7ca0>
      name = <local> 'pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.utils.rnn.pack_padded_sequence(...)', len = 92
      ctx = <local> <WrapCtx 'pytorch_to_returnn.import_wrapper._torch_traced'>
  File "/u/vieting/testing/20210714_dfg_kickoff/pytorch-to-returnn/pytorch_to_returnn/import_wrapper/base_wrappers/object.py", line 88, in make_wrapped_object
    line: assert issubclass(cls, WrappedClassBase)
    locals:
      issubclass = <builtin> <built-in function issubclass>
      cls = <local> <class 'torch.nn.utils.rnn.PackedSequence'>
      WrappedClassBase = <local> <class 'pytorch_to_returnn.import_wrapper.base_wrappers.class_.WrappedClassBase'>
AssertionError
```

pytorch_to_returnn/import_wrapper/wrap.py
```diff
@@ -38,6 +38,8 @@ def wrap(obj, *, name: str, ctx: WrapCtx):
       pass  # keep as-is
     elif type(obj) != type:  # explicitly do not check sub-types, e.g. pybind11_type
       pass
+    elif issubclass(obj, tuple):  # e.g. namedtuple
+      pass  # keep as-is
```
albertz, line 42: Why?

pytorch_to_returnn/import_wrapper/wrap.py
albertz: You still have this here after our recent changes. Is this still needed?

pytorch_to_returnn/torch/nn/modules/rnn.py
```diff
@@ -11,12 +12,42 @@
 from ..parameter import Parameter
 from ...tensor import Tensor
 from .. import init
+from ....naming import Naming, TensorEntry
 
 
-class PackedSequence:  # dummy -- not yet implemented...
+class BatchSizesDummy:
```
albertz, line 18: Do you still need this?

pytorch_to_returnn/torch/nn/modules/rnn.py
```diff
@@ -11,12 +12,42 @@
...
+  def get_batched_tensor(self) -> Tensor:
+    from .shape import FlattenBatch
+    naming = Naming.get_instance()
+    tensor_entry = naming.tensors[self.data]
+    assert isinstance(tensor_entry, TensorEntry)
+    for call_entry in tensor_entry.output_from_calls:
+      if isinstance(call_entry.module.module, FlattenBatch):
+        return call_entry.orig_inputs_args[0]
+    from .shape import UnflattenBatch
+    return UnflattenBatch()(self.data)
```
albertz, line 48: Maybe we want to throw an exception and not do the `UnflattenBatch` here? Not sure...

pytorch_to_returnn/torch/nn/utils/rnn.py
```diff
@@ -0,0 +1,32 @@
...
+def pad_packed_sequence(sequence: PackedSequence, batch_first=False, padding_value=0.0, total_length=None):
+  assert padding_value == 0.0, "not implemented yet"
+  assert total_length is None, "not implemented yet"
+
+  return sequence.get_batched_tensor(), sequence.batch_sizes or BatchSizesDummy()
+
+
+def pack_sequence(sequences, enforce_sorted=True):
+  assert enforce_sorted, "not implemented yet"
+  if isinstance(sequences, Tensor):
```
albertz, line 20: Could it ever be not a `Tensor` here?

pytorch_to_returnn/torch/nn/utils/rnn.py
```diff
@@ -0,0 +1,32 @@
...
+
+def pad_packed_sequence(sequence: PackedSequence, batch_first=False, padding_value=0.0, total_length=None):
+  assert padding_value == 0.0, "not implemented yet"
+  assert total_length is None, "not implemented yet"
+
+  return sequence.get_batched_tensor(), sequence.batch_sizes or BatchSizesDummy()
+
+
+def pack_sequence(sequences, enforce_sorted=True):
+  assert enforce_sorted, "not implemented yet"
```
albertz, line 19: We actually do not enforce that it is sorted. But we ignore it anyway. I would just remove this.

pytorch_to_returnn/torch/nn/utils/rnn.py
albertz: Ah, I guess yes, otherwise there is no real difference to `pack_padded_sequence`, right?

pytorch_to_returnn/torch/nn/utils/rnn.py
```diff
@@ -0,0 +1,32 @@
...
+  if isinstance(sequences, Tensor):
+    naming = Naming.get_instance()
+    tensor_entry = naming.tensors[sequences]
+    if tensor_entry.returnn_data.have_batch_axis():
```
albertz, line 23: Maybe remove this check, and leave the logic to `FlattenBatch`.

pytorch_to_returnn/torch/nn/utils/rnn.py
```diff
@@ -0,0 +1,32 @@
...
+      return PackedSequence(FlattenBatch()(sequences))
```
albertz, line 24: Better call `pack_padded_sequence` here. This would also have the different `batch_major` default.

albertz:
Should we later squash all the commits together? It involves some logical separated changes. If squashing all together is not wanted, you should cleanup the history, and manually squash them together as far as it make sense.

pytorch_to_returnn/torch/nn/modules/rnn.py
vieting: That's what you initially proposed, but I added this as a fallback. In the example that I'd like to convert, there is something like 

`h = pack_sequence(...); h_data = some_op(h.data); h = PackedSequence(h_data, h.batch_sizes)`

Then, `self.data` is not a direct output of `FlattenBatch`, so it cannot be found in the way above.

pytorch_to_returnn/torch/nn/utils/rnn.py
vieting: Yes, actually it should not be a `Tensor` but a `list[Tensor]`, see https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.pack_sequence.html

But it also works with a batched tensor.

pytorch_to_returnn/import_wrapper/wrap.py
vieting: To allow unpacking, e.g. `data, batch_sizes, sorted_indices, unsorted_indices = packed_seq`.

vieting:
Yes, I'll cleanup the history.

pytorch_to_returnn/import_wrapper/wrap.py
albertz: I don't understand:

- Why or in what situation do we even call this function for a `PackedSequence`?
- Why does the unpacking not work for the wrapped object? This sounds like a bug which would be easy to fix, and which should be fixed anyway, and then this change here is not needed.
- Wouldn't this cause problems for other named tuples which are not over tensors?
- Doesn't this cause problems when the tensors inside the namedtuple are not wrapped then?

vieting: > Why or in what situation do we even call this function for a PackedSequence?

This is called in `_run_traced_orig_torch()` when a `PackedSequence` is input to some module, like `LSTM` in `test_lstm_with_packed_sequence_input`. For the `LSTM` module, we run into a problem if the class is wrapped, because in the `forward`, there is a check `if isinstance(orig_input, PackedSequence): input, batch_sizes, sorted_indices, unsorted_indices = input`. For the wrapped object, the unpacking is skipped which is why the subsequent code crashes.

> Why does the unpacking not work for the wrapped object? This sounds like a bug which would be easy to fix, and which should be fixed anyway, and then this change here is not needed.

Actually, the unpacking works without the marked change. It was probably some other issue before.

> Wouldn't this cause problems for other named tuples which are not over tensors?
> Doesn't this cause problems when the tensors inside the namedtuple are not wrapped then?

This here is also not only over tensors, but so far always includes `None`. I didn't encounter issues because of non-wrapped tensors. In what situation could that occur?

albertz: > > Why or in what situation do we even call this function for a PackedSequence?
> 
> This is called in `_run_traced_orig_torch()` when a `PackedSequence` is input to some module, like `LSTM` in `test_lstm_with_packed_sequence_input`.

Ah, I see now.

> For the `LSTM` module, we run into a problem if the class is wrapped, because in the `forward`, there is a check `if isinstance(orig_input, PackedSequence): input, batch_sizes, sorted_indices, unsorted_indices = input`.

Ah, but this can be fixed. We just also need to implement `__instancecheck__` and maybe `__subclasscheck__` in the wrapped object (in `WrappedObject`).

Also, isn't `PackedSequence` here also itself wrapped (via `make_wrapped_class`)? If not, why not? If yes, then why doesn't it work already? In `make_wrapped_object`, it would specifically create a wrapped object of the right type, so in principle, this `isinstance` check should already work.

> > Why does the unpacking not work for the wrapped object? This sounds like a bug which would be easy to fix, and which should be fixed anyway, and then this change here is not needed.
> 
> Actually, the unpacking works without the marked change. It was probably some other issue before.
> 
> > Wouldn't this cause problems for other named tuples which are not over tensors?
> > Doesn't this cause problems when the tensors inside the namedtuple are not wrapped then?
> 
> This here is also not only over tensors, but so far always includes `None`. I didn't encounter issues because of non-wrapped tensors. In what situation could that occur?

When you don't wrap `PackedSequence` (or other tuples or objects over tensors), then it would also not wrap the tensor inside. At least not here. Maybe it would wrap it by some follow-up code when the tensor is directly used. But not sure. Or maybe it doesn't cause problems here because in our Torch implementation, it would not exist like this.

However, this is now very specific to `PackedSequence`. In general, when a tuple or namedtuple contains tensors, those should be wrapped.

What we could do is to not wrap tuple, namedtuple etc but unpack them, wrap all their content, and then repack them. However, I'm not sure if this is better, or really needed. In principle, wrapping a tuple directly should also work.

vieting: > Also, isn't PackedSequence here also itself wrapped (via make_wrapped_class)? If not, why not? If yes, then why doesn't it work already? In make_wrapped_object, it would specifically create a wrapped object of the right type, so in principle, this isinstance check should already work.

Yes, as I wrote in the edit above: In the padertorch example it works and I'm not sure why there is this difference. In the padertorch example, this function gets called for `PackedSequence` already a couple of times when padertorch is imported, maybe this makes the difference.

pytorch_to_returnn/torch/nn/modules/module.py
```diff
@@ -623,15 +626,20 @@ def check_call_returnn_outputs_to_prev_torch(call: CallEntry, *, update_ops: Lis
       idx_repr = f" {i + 1}/{len(call.outputs_flat)}" if len(call.outputs_flat) > 1 else ""
       print(f"**** validate: add call {call} output{idx_repr} tensor {x}")
     out, _ = session.run((
-      [(x.returnn_data.placeholder, x.returnn_data.size_placeholder.as_dict()) for x in call.outputs_flat], update_ops),
+      [(x.returnn_data.placeholder, x.returnn_data.size_placeholder.as_dict()) if x is not None else tf.constant(0) for
```
vieting, line 629:
```suggestion
      [(x.returnn_data.placeholder, x.returnn_data.size_placeholder.as_dict()) if x is not None else {} for
```

We could do this change to avoid unnecessary ops in the graph, right?

pytorch_to_returnn/import_wrapper/wrap.py
albertz: > but then the unpacking fails (`TypeError: cannot unpack non-iterable PackedSequence object`).

So `isinstance(orig_input, PackedSequence)` returned True, and then the unpacking fails with this error? It's strange that it works in some cases but not in others.

But in any case, the unpacking of a `WrappedObject` is probably just not implemented right. I assume the dynamic `__getattr__` logic does not work, i.e. will not be called for other magic methods like `__len__`, `__iter__`, etc.

So maybe we should add such code to `WrappedObject.__init__`:
```
for special in ["__len__", "__iter__"]:
  if hasattr(orig_obj, special):
    setattr(self, getattr(self, special))  # make explicitly available
```

pytorch_to_returnn/torch/nn/modules/module.py
albertz: Ah yes, does that work? If it works, then sure, much better. Or maybe empty tuple instead.

vieting: Yes, both work.

pytorch_to_returnn/import_wrapper/wrap.py
```diff
@@ -77,6 +81,8 @@ def unwrap(obj):
 def _nested_transform(obj, transform):
   if type(obj) == tuple:
     return tuple([transform(item) for item in obj])
+  if issubclass(type(obj), tuple) and type(obj) is not tuple:
+    return type(obj)(*(transform(item) for item in obj))
```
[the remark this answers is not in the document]
vieting: I suppose so, I'll remove it.

pytorch_to_returnn/import_wrapper/wrap.py
vieting: It seems like this doesn't change anything. But the problem seems to be related. At least, I cannot call `len()` on the `PackedSequence`

```
>>> len(input)
*** TypeError: object of type 'PackedSequence' has no len()
>>> input.__len__()
4
```

pytorch_to_returnn/import_wrapper/wrap.py
vieting: Yes, this is still needed. Without it, `make_wrapped_object` fails for a `PackedSequence` in `_run_traced_orig_torch` with the same error as above.

pytorch_to_returnn/import_wrapper/wrap.py
albertz: Ah yes, it doesn't work like that. It needs to be already assigned on the class, not on the object.

I think you need to add similar logic to `make_wrapped_class`. Sth like:
```
for special in ["__len__", "__iter__"]:
  if hasattr(cls, special):
    func = getattr(cls, special)
    wrapped_func = wrap(func, name="%s.%s" % (name, special), ctx=ctx)
    setattr(WrappedClass, wrapped_func)  # make explicitly available
```

pytorch_to_returnn/import_wrapper/wrap.py
albertz: But why? Why is it not wrapped? Or is it, and the type is messed up? Can you check `inspect.getmro(cls)`?

Or is this just together with the other change above, where you don't wrap types with `issubclass(obj, tuple)`? The error is gone when you remove both?

pytorch_to_returnn/import_wrapper/wrap.py
vieting: Yes, this seems to work with some small modifications :+1:

vieting: What remains are these points

> > For the `LSTM` module, we run into a problem if the class is wrapped, because in the `forward`, there is a check `if isinstance(orig_input, PackedSequence): input, batch_sizes, sorted_indices, unsorted_indices = input`.
>
> Ah, but this can be fixed. We just also need to implement `__instancecheck__` and maybe `__subclasscheck__` in the wrapped object (in `WrappedObject`).
>
> Also, isn't `PackedSequence` here also itself wrapped (via `make_wrapped_class`)? If not, why not? If yes, then why doesn't it work already? In `make_wrapped_object`, it would specifically create a wrapped object of the right type, so in principle, this `isinstance` check should already work.
>
> > > Why does the unpacking not work for the wrapped object? This sounds like a bug which would be easy to fix, and which should be fixed anyway, and then this change here is not needed.
> > 
> > Actually, the unpacking works without the marked change. It was probably some other issue before.
> > 
> > > Wouldn't this cause problems for other named tuples which are not over tensors?
> > > Doesn't this cause problems when the tensors inside the namedtuple are not wrapped then?
> > 
> > This here is also not only over tensors, but so far always includes `None`. I didn't encounter issues because of non-wrapped tensors. In what situation could that occur?
>
> When you don't wrap `PackedSequence` (or other tuples or objects over tensors), then it would also not wrap the tensor inside. At least not here. Maybe it would wrap it by some follow-up code when the tensor is directly used. But not sure. Or maybe it doesn't cause problems here because in our Torch implementation, it would not exist like this.
>
> However, this is now very specific to `PackedSequence`. In general, when a tuple or namedtuple contains tensors, those should be wrapped.
>
> What we could do is to not wrap tuple, namedtuple etc but unpack them, wrap all their content, and then repack them. However, I'm not sure if this is better, or really needed. In principle, wrapping a tuple directly should also work.

vieting: The difference is that in the padertorch example, we have something like

```
h = torch.nn.utils.rnn.pack_padded_sequence(inputs, [n_time] * n_batch)
...
h = PackedSequence(h.data, h.batch_sizes)
output, _ = torch.nn.LSTM(n_feat, n_feat)(h)
```

while in our test case, there is no direct initialization of the `PackedSequence`. This way, in our test `h` is of the original type, not the wrapped one.

pytorch_to_returnn/import_wrapper/wrap.py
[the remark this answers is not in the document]
albertz: Another thing, which should maybe be at the top of this function:
```
if isinstance(obj, WrappedObject):
  return obj
```
Otherwise it would keep the wrapped container type (namedtuple, `PackedSequence`) but unwrap all its content (the tensors), and then it might be confused in other parts? Not sure... But it should also not cause problems, and be safe, and better to have this.

pytorch_to_returnn/import_wrapper/wrap.py
vieting: Oh yes, it might be an interdependency of the two things. Now I get some other error, looking into this...

vieting: If we have a `PackedSequence` as input to a module (here `LSTM`), there is a mismatch in the flattened inputs for `_run_traced_orig_torch()` and `_run_torch_returnn_drop_in()`. In the first case, the input is of type `pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.utils.rnn.PackedSequence` which does not get flattened by `nest.flatten()`. In the second case, the input is of type `pytorch_to_returnn.torch.nn.modules.rnn.PackedSequence` which gets flattened into `[Tensor, None, None, None]`. Later, a check that both flattened inputs have the same lengths will fail.

albertz: > `pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.utils.rnn.PackedSequence` which does not get flattened by `nest.flatten()`

Why?

vieting: I suppose because `pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.utils.rnn.PackedSequence` is not a subclass of `tuple` and this is because `WrappedClass.__bases__ += (cls,)` in `make_wrapped_class()` throws an error which is ignored by the following exception (`TypeError: __bases__ assignment: 'PackedSequence' object layout differs from 'WrappedClassBase'`).

vieting:
I went back to not wrapping `namedtuple`s. Some checks did not work as I had in mind, because the `batch_sizes` were not set correctly. Therefore, I updated this part so that they should be correct now. We just assume that they are set correctly in the RETURNN `size_placeholder` already in the input to the converter.

pytorch_to_returnn/import_wrapper/base_wrappers/object.py
```diff
@@ -13,6 +13,13 @@ def __init__(self, orig_obj: Any, name: str, ctx: WrapCtx):
     self._wrapped__orig_obj = orig_obj
     self._wrapped__ctx = ctx
 
+    for special in ["__len__", "__iter__"]:
+      if hasattr(orig_obj, special):
+        from ..wrap import wrap
+        func = getattr(orig_obj, special)
+        wrapped_func = wrap(func, name="%s.%s" % (name, special), ctx=ctx)
+        setattr(type(self), special, wrapped_func)  # make explicitly available
```
albertz, line 21: This should not be here in but in the wrapped class.

pytorch_to_returnn/import_wrapper/wrap.py
albertz: This here should also have a small comment. Or maybe just refer to the comment below.

albertz (approved):

PR #72, opened 2021-12-01
fix batch_first in pack_sequence

vieting:
Small fix following up on #71.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/rnn.py
+++ b/pytorch_to_returnn/torch/nn/modules/rnn.py
@@ -29,16 +29,22 @@ class PackedSequence(PackedSequence_):
   always be satisfied.
   """
 
-  def get_batched_tensor(self) -> Tensor:
+  def get_padded_tensor(self, batch_first: bool) -> Tensor:
     from .shape import FlattenBatch
     naming = Naming.get_instance()
     tensor_entry = naming.tensors[self.data]
     assert isinstance(tensor_entry, TensorEntry)
     for call_entry in tensor_entry.output_from_calls:
       if isinstance(call_entry.module.module, FlattenBatch):
-        return call_entry.orig_inputs_args[0]
+        out_tensor = call_entry.orig_inputs_args[0]
+        if batch_first != out_tensor.shape[0].is_batch_dim:
+          # batch_first refers to whether the output should be batch major or not. Since the input to the PackedSequence
+          # was not batch major in this case, we need to transpose.
+          out_tensor = out_tensor.transpose(0, 1)
+        assert batch_first == out_tensor.shape[0].is_batch_dim
+        return out_tensor
     from .shape import UnflattenBatch
-    return UnflattenBatch()(self.data)
+    return UnflattenBatch(batch_first=batch_first)(self.data)
 
 
 def apply_permutation(tensor: Tensor, permutation: Tensor, dim: int = 1) -> Tensor:
@@ -206,7 +212,7 @@ class LSTM(RNNBase):
 
   def create_returnn_layer_dict(self, input: Tensor, hx: Optional[Tensor] = None) -> Dict[str, Any]:
     if isinstance(input, PackedSequence):
-      input = input.get_batched_tensor()
+      input = input.get_padded_tensor(batch_first=False)
     if not self.bidirectional and self.num_layers == 1:
       d = {
         "class": "rec", "unit": "nativelstm2", "from": self._get_input_layer_name(input),
@@ -257,7 +263,7 @@ class LSTM(RNNBase):
     assert call_entry.module.module is self
     orig_input = call_entry.orig_inputs_args[0]
     if isinstance(orig_input, PackedSequence):
-      orig_input = orig_input.get_batched_tensor()
+      orig_input = orig_input.get_padded_tensor(batch_first=False)
     return self._base_get_output_shape_from_returnn(inputs_flat=[orig_input], layer=layer)
 
   def make_structured_returnn_output(self, output: Tensor, input: Tensor, hx: Optional[Tensor] = None
--- a/pytorch_to_returnn/torch/nn/modules/shape.py
+++ b/pytorch_to_returnn/torch/nn/modules/shape.py
@@ -368,8 +368,9 @@ class FlattenBatch(Module):
 class UnflattenBatch(Module):
   is_original_torch_module = False
 
-  def __init__(self):
+  def __init__(self, batch_first: bool):
     super(UnflattenBatch, self).__init__()
+    self.batch_first = batch_first
 
   def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
     return {"class": "unflatten_batch", "from": self._get_input_layer_name(input)}
@@ -382,6 +383,12 @@ class UnflattenBatch(Module):
     assert x.shape[0].merged_dims
     torch_shape = tuple(x.shape[0].merged_dims) + x.shape[1:]
     returnn_axis_from_torch_axis = {i: i for i in range(len(torch_shape))}
+    if self.batch_first != torch_shape[0].is_batch_dim:
+      torch_shape = (torch_shape[1], torch_shape[0]) + torch_shape[2:]
+    if self.batch_first != layer.output.is_batch_major:
+      # batch and unflattened axis are always the first two, so just swap them
+      returnn_axis_from_torch_axis[0] = 1
+      returnn_axis_from_torch_axis[1] = 0
     return torch_shape, returnn_axis_from_torch_axis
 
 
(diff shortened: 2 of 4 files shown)


pytorch_to_returnn/torch/nn/utils/rnn.py
```diff
@@ -37,7 +37,8 @@ def pack_sequence(sequences, enforce_sorted=True):
     batch_dim = sequences.shape[torch_axis_from_returnn_axis[tensor_entry.returnn_data.batch_dim_axis]]
     time_dim = sequences.shape[torch_axis_from_returnn_axis[tensor_entry.returnn_data.time_dim_axis]]
     lengths = [time_dim] * batch_dim
-    return pack_padded_sequence(sequences, lengths, enforce_sorted=enforce_sorted)
+    batch_first = True  # we assume batch-time-major tensor, see above
```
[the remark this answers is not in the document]
vieting: I think the `batch_first` in `pack_padded_sequence` refers to its input. And here, the input is always batch major, because only in this case `pack_sequence` with a batched tensor works as expected.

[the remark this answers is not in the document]
albertz: Oh really? Then we handle it incorrectly in `pack_padded_sequence`. Because we use it for the output, i.e. the packed tensor. Or maybe in PyTorch it refers to both input and output? Then it would be correct because the input should not matter at all. It should know internally what the batch dim is and handle any case correctly.

In any case, the input should not matter here at all and should not make any difference. This just has an effect on the output. And I wonder, in all usages of `PackedSequence` in PyTorch that I know (`LSTM` etc), it expects a time-major packed tensor.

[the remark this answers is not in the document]
vieting: I think in `pack_padded_sequence` it refers to whether the input is a batch major tensor and in `pad_packed_sequence` it refers to whether the output should be batch major or not.

The packed tensor looks the same whether `batch_first` is `True` or `False` in  `pack_padded_sequence`. So I guess `FlattenBatch` should always use `batch_first=False`. For `pad_packed_sequence` we need to extend a bit, because PyTorch returns it according to `batch_first` (by default `False`) but our `PackedSequence.get_batched_tensor` returns it in whatever format the input before flattening was.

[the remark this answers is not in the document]
vieting: Do you also think so?

vieting:
I added two tests which cover the cases which are currently not handled correctly: 1) packing the tensor with `batch_first=True` in `FlattenBatch` and then directly accessing the packed tensor and 2) packing the tensor with `batch_first=False` but unpacking with `batch_first=True`.

pytorch_to_returnn/torch/nn/utils/rnn.py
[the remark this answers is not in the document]
albertz: `FlattenBatch` should always use `batch_major=False` then, yes.

And I think we can just ignore the `batch_first`. Or we can make an assertion that it matches what we have in the Torch tensor.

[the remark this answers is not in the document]
vieting: I think we can take care of `batch_first` in `get_batched_tensor`, see the most recent commit. Note that the initial change in this PR is removed then.

pytorch_to_returnn/torch/nn/modules/rnn.py
```diff
@@ -29,16 +29,25 @@ class PackedSequence(PackedSequence_):
   always be satisfied.
   """
 
-  def get_batched_tensor(self) -> Tensor:
+  def get_batched_tensor(self, batch_first=None) -> Tensor:
```
albertz, line 32: Maybe we should rename this function to `get_padded_tensor` to make it more clear.

albertz: You should also specify the type of `batch_first` (`Optional[bool]`).

pytorch_to_returnn/torch/nn/modules/rnn.py
```diff
@@ -29,16 +29,25 @@ class PackedSequence(PackedSequence_):
...
-        return call_entry.orig_inputs_args[0]
+        out_tensor = call_entry.orig_inputs_args[0]
+        if batch_first is not None:
+          if batch_first != out_tensor.shape[0].is_batch_dim:
+            out_tensor = out_tensor.transpose(0, 1)
+          out_tensor_entry = naming.tensors[out_tensor]
+          if batch_first != out_tensor_entry.returnn_data.is_batch_major:
+            # batch and unflattened axis are always the first two, so just swap them
+            out_tensor_entry.returnn_axis_from_torch_axis[0] = 1
+            out_tensor_entry.returnn_axis_from_torch_axis[1] = 0
```
albertz, line 47: I don't understand this.

First of all, the check is wrong. It never matters whether the RETURNN data is batch major. It matters whether the Torch tensor is batch major. But you already have that in your check above.

Then you cannot just reassign `returnn_axis_from_torch_axis`. Call `transpose` or so instead. But actually, again, this is just the code you already have above?

I think you don't need anything here anymore. Maybe just a check like:
```suggestion
          assert batch_first == out_tensor.shape[0].is_batch_dim
```

pytorch_to_returnn/torch/nn/modules/rnn.py
albertz: Btw, do you want to allow `batch_first` to be `None` here? Where do we call this `get_batched_tensor` (`get_padded_tensor`)? The resulting Torch tensor, maybe we always want to have it explicit whether it is batch major or not?

pytorch_to_returnn/torch/nn/modules/shape.py
```diff
@@ -368,8 +368,9 @@ def _get_output_shape_from_returnn(self,
 class UnflattenBatch(Module):
   is_original_torch_module = False
 
-  def __init__(self):
+  def __init__(self, batch_first: Optional[bool] = None):
```
albertz, line 371: Also here, maybe it should not be optional. Maybe this should always be explicit. Because on Torch side, it will always matter which way it is.

pytorch_to_returnn/torch/nn/utils/rnn.py
```diff
@@ -14,14 +14,14 @@ def pack_padded_sequence(input: Tensor, lengths, batch_first=False, enforce_sort
 
 
 def pack_padded_sequence_with_batch_sizes(input: Tensor, batch_sizes: Tensor, batch_first=False) -> PackedSequence:
-  return PackedSequence(FlattenBatch(batch_major=batch_first)(input), batch_sizes)
+  return PackedSequence(FlattenBatch(batch_major=False)(input), batch_sizes)
```
albertz, line 17: You should add some comment explaining what we discussed, i.e. that `PackedSequence` always uses the `batch_major=False` flattened variant. And that `batch_first` is about the input here, which we don't need to use as we keep that information internally around anyway.

But also, here you might want to some extra check, like:
```suggestion
  assert batch_first == input.shape[0].is_batch_dim
  return PackedSequence(FlattenBatch(batch_major=False)(input), batch_sizes)
```

pytorch_to_returnn/torch/nn/modules/rnn.py
vieting: The renaming makes it also more consistent with the PyTorch names. We call this function in `pad_packed_sequence` and in the `LSTM` module. I'm fine with making it always explicit, this is probably clearer.

pytorch_to_returnn/torch/nn/modules/rnn.py
vieting: Ah yes, you're right. I removed this now.

pytorch_to_returnn/torch/nn/modules/shape.py
vieting: The advantage of allowing `None` would be that we would get the order in RETURNN that was used before flattening. But yes, on Torch side it's always explicit so I'm fine with being explicit here as well.

pytorch_to_returnn/torch/nn/utils/rnn.py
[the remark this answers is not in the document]
vieting: I added this again, as `batch_first` refers to the input which is a batch major tensor here, so `batch_first` should be `True`. (It's not used but we added the assertion in `pack_padded_sequence_with_batch_sizes` now).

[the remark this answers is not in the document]
albertz: I would not introduce it as a local variable put just pass `batch_first=batch_first` directly below. But with the same comment before that we assume a batch-major Torch tensor here.

PR #74, opened 2021-12-02
Add permute

vieting:
Add `torch.nn.functional.permute` and `Tensor.permute` which uses the previous one. In order to do this, I also fixed `out_returnn_axis_from_torch_axis` for `Transpose`. Apparently, we only had cases where two axes were transposed, not where a more complicated transpose was done like 0->1, 1->2, 2->0, so the error did not become obvious before.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -260,9 +260,10 @@ class Transpose(Module):
     assert tensor_entry.returnn_data and tensor_entry.returnn_axis_from_torch_axis
     assert tensor_entry.returnn_data.batch_ndim == ndim
 
+    inv_perm = {j: i for i, j in perm.items()}
     out_torch_shape = [input.shape[perm[i]] for i in range(ndim)]
     out_returnn_axis_from_torch_axis = {
-      perm[i]: j for (i, j) in tensor_entry.returnn_axis_from_torch_axis.items()}
+      inv_perm[i]: j for (i, j) in tensor_entry.returnn_axis_from_torch_axis.items()}
     return tuple(out_torch_shape), out_returnn_axis_from_torch_axis
 
 
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -1033,6 +1033,18 @@ def test_movedim_2():
   verify_torch_and_convert_to_returnn(model_func, inputs=x)
 
 
+def test_permute():
+  n_batch, n_time, n_feature = 3, 5, 7
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    out = inputs.permute(2, 0, 1)
+    return out
+
+  rnd = numpy.random.RandomState(42)
+  x = rnd.normal(0., 1., (n_batch, n_feature, n_time)).astype("float32")
+  verify_torch_and_convert_to_returnn(model_func, inputs=x)
+
+
 def test_flatten_batch():
   n_in, n_out = 11, 13
   n_batch, n_time = 3, 7
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -1017,6 +1017,22 @@ def test_movedim():
   verify_torch_and_convert_to_returnn(model_func, inputs=x)
 
 
+def test_movedim_2():
+  n_batch, n_time, n_feature = 3, 5, 7
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    if typing.TYPE_CHECKING or not wrapped_import:
+      import torch
+    else:
+      torch = wrapped_import("torch")
+    out = torch.movedim(inputs, 0, 2)
+    return out
+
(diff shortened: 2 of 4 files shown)


pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -262,7 +262,7 @@ def _get_output_shape_from_returnn(self,
 
     out_torch_shape = [input.shape[perm[i]] for i in range(ndim)]
     out_returnn_axis_from_torch_axis = {
-      perm[i]: j for (i, j) in tensor_entry.returnn_axis_from_torch_axis.items()}
+      j: perm[i] for (i, j) in tensor_entry.returnn_axis_from_torch_axis.items()}
```
albertz, line 265: This looks wrong.

Actually, the original code looks correct. Why do you think it is wrong?

Just the name `returnn_axis_from_torch_axis` implies that the keys are the Torch axes, while the values are the RETURNN axes. You are swapping keys and values here in your code, which would be wrong.

The `perm` should be applied on the Torch axes. I.e. on the keys. Which is what was done before.

tests/test_layers.py
```diff
@@ -1017,6 +1017,34 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    if typing.TYPE_CHECKING or not wrapped_import:
+      import torch
+    else:
+      torch = wrapped_import("torch")
+    out = torch.movedim(inputs, 0, 2)
+    return out
+
+  rnd = numpy.random.RandomState(42)
+  x = rnd.normal(0., 1., (n_batch, n_time, n_feature)).astype("float32")
```
albertz, line 1032: Although I think it doesn't really matter here, the order is wrong.
```suggestion
  x = rnd.normal(0., 1., (n_batch, n_feature, n_time)).astype("float32")
```

tests/test_layers.py
```diff
@@ -1017,6 +1017,34 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
+
+def test_permute():
+  n_batch, n_time, n_feature = 3, 5, 7
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    out = inputs.permute(2, 0, 1)
+    return out
+
+  rnd = numpy.random.RandomState(42)
+  x = rnd.normal(0., 1., (n_batch, n_time, n_feature)).astype("float32")
```
albertz, line 1044: Same here.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: Hm, that's right. But I added the `test_movedim_2` to show that the current version fails in that case. I think then we need to take `perm` of `j` instead of `i`

```suggestion
      i: perm[j] for (i, j) in tensor_entry.returnn_axis_from_torch_axis.items()}
```

vieting: This also works. I think it is even equivalent with my suggestion above (`i: perm[j]`)

albertz: Ah yes I think you are right. This always confuses me.

tests/test_layers.py
```diff
@@ -1017,6 +1017,34 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    if typing.TYPE_CHECKING or not wrapped_import:
+      import torch
+    else:
+      torch = wrapped_import("torch")
+    out = torch.movedim(inputs, 0, 2)
+    return out
+
+  rnd = numpy.random.RandomState(42)
+  x = rnd.normal(0., 1., (n_time, n_batch, n_feature)).astype("float32")
```
albertz, line 1032: Still wrong order? Shouldn't it be:
```suggestion
  x = rnd.normal(0., 1., (n_batch, n_feature, n_time)).astype("float32")
```

tests/test_layers.py
```diff
@@ -1017,6 +1017,34 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
+
+def test_permute():
+  n_batch, n_time, n_feature = 3, 5, 7
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    out = inputs.permute(2, 0, 1)
+    return out
+
+  rnd = numpy.random.RandomState(42)
+  x = rnd.normal(0., 1., (n_time, n_batch, n_feature)).astype("float32")
```
albertz, line 1044: Same here.

tests/test_layers.py
vieting: Yes, you're right, sorry.

albertz (approved):

albertz:
This is ready, right? Or is there any reason this is still in draft state?

vieting:
Yes, it's ready.

PR #76, opened 2022-01-13
add torch.sum and Tensor.sum

vieting:
Use RETURNN's `ReduceLayer` to add `torch.sum` and `Tensor.sum`.

### DIFF
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -124,6 +124,11 @@ def is_tensor(obj) -> bool:
   return isinstance(obj, Tensor)
 
 
+def sum(input: Tensor, dim: Optional[int] = None, dtype: Optional[Union[str, _dtype]] = None) -> Tensor:
+  assert dim is not None, "not implemented yet"
+  return modules.Reduce(mode="sum", axes=dim)(input)
+
+
 def add(x: Tensor, y: Tensor) -> Tensor:
   dtype = result_type(x, y)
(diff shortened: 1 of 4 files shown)


vieting:
I just saw that I'm overriding the inbuilt `sum` in `functional` which is not a good idea. I don't actually need to have it in `functional`, but in PyTorch it's available as `torch.sum` and we usually have these functions in `functional.py`. Where should we put this here instead?

albertz:
> I just saw that I'm overriding the inbuilt `sum` in `functional` which is not a good idea. I don't actually need to have it in `functional`, but in PyTorch it's available as `torch.sum` and we usually have these functions in `functional.py`. Where should we put this here instead?

I don't understand the problem. Wherever you define it, you would overwrite the `sum` builtin. Why is it a problem specifically in `functional`? And why is overwriting the builtin `sum` a problem? If you need the builtin `sum` elsewhere in the module, you can simply do:
```
_builtin_sum = sum

...

def sum(...):
  ...
```

albertz (approved):
Looks good. Have you tested it?

albertz:
I wonder whether we should first fix the issue with RETURNN (#78).

vieting:
> I wonder whether we should first fix the issue with RETURNN (#78), and then rebase here to check that all tests run through.

Yes, that was my plan.

albertz:
I think this needs a rebase now for the new `reuse_params` logic.

vieting:
>I think this needs a rebase now for the new reuse_params logic.

It's now rebased and all tests pass.

PR #77, opened 2022-01-13
allow gathering with tensors

vieting:
I would like to allow gather a tensor with another tensor which contains integer indices which the `GatherLayer` should already support.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -273,18 +273,34 @@ class Gather(Module):
   """
   is_original_torch_module = False
 
-  def __init__(self, dim: int, pos: Union[int, Tensor]):
+  def __init__(self, dim: int, pos: int):
     super(Gather, self).__init__()
     self.dim = dim
     self.pos = pos
 
   def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
-    if isinstance(self.pos, Tensor):
-      self.pos = self._get_input_layer_name(self.pos)
     return {
       "class": "gather", "from": self._get_input_layer_name(input),
       "axis": self._get_input_axis_to_returnn(input, axis=self.dim),
       "position": self.pos}
+
+
+class GatherTensor(Module):
+  """
+  Basically x[pos] but in specific dim (axis) where pos is a tensor.
+  """
+  is_original_torch_module = False
+
+  def __init__(self, dim: int):
+    super(GatherTensor, self).__init__()
+    self.dim = dim
+
+  def create_returnn_layer_dict(self, input: Tensor, pos: Tensor) -> Dict[str, Any]:
+    pos = self._get_input_layer_name(pos)
+    return {
+      "class": "gather", "from": self._get_input_layer_name(input),
+      "axis": self._get_input_axis_to_returnn(input, axis=self.dim),
+      "position": pos}
 
 
 class Slice(Module):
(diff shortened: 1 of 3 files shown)


vieting:
In the test case that I added, this works already. However, in the case that I'm actually interested in, the `GatherLayer` returns a wrong shape and I currently cannot reproduce this in a test case. There, the input is (B&Packed|57, F|768), the indices for slicing (F|5700) but the output is (F|768) instead of (5700, F|768).

Do you see some pitfalls which have to be considered here?

vieting:
Btw, I think the failing test case is not related to the changes introduced in this PR, probably due to updates in RETURNN.

vieting:
It's now rebased and ready for review. The issue that I mentioned above was also solved by the RETURNN related fixes.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -273,12 +273,14 @@ class Gather(Module):
...
 
-  def __init__(self, dim: int, pos: int):
+  def __init__(self, dim: int, pos: Union[int, Tensor]):
     super(Gather, self).__init__()
     self.dim = dim
     self.pos = pos
 
   def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
+    if isinstance(self.pos, Tensor):
+      self.pos = self._get_input_layer_name(self.pos)
```
albertz, line 283: You should not reassign the attrib here. Use a local var.

albertz (changes requested):
Small change

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -273,12 +273,14 @@ class Gather(Module):
   """
   is_original_torch_module = False
 
-  def __init__(self, dim: int, pos: int):
+  def __init__(self, dim: int, pos: Union[int, Tensor]):
```
albertz, line 276: I'm not sure but maybe it's a bad idea (inconsistent) to have tensors as arguments in the `__init__`?

albertz: Maybe it makes more sense to have a separate module (`Slice` or so) for this case, where `pos` is then an argument for `create_returnn_layer_dict` and not for `__init__`.

vieting: I pushed this now.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -285,6 +285,24 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
...
+  def __init__(self, dim: int):
+    super(GatherTensor, self).__init__()
+    self.dim = dim
+
+  def create_returnn_layer_dict(self, input: Tensor, pos: Tensor) -> Dict[str, Any]:
+    pos = self._get_input_layer_name(pos)
+    return {
+      "class": "gather", "from": self._get_input_layer_name(input),
+      "axis": self._get_input_axis_to_returnn(input, axis=self.dim),
+      "position": pos}
```
albertz, line 303: I think this looks a bit more consistent:
```suggestion
    return {
      "class": "gather", "from": self._get_input_layer_name(input),
      "axis": self._get_input_axis_to_returnn(input, axis=self.dim),
      "position": self._get_input_layer_name(pos)}
```

albertz (approved):

PR #80, opened 2022-01-18
fix Unflatten out_returnn_axis_from_torch_axis

vieting:
The `elif` branch that I edited is supposed to be the case where `i` refers to a dim which outside (larger than) the unflattened dims. Equality should also be contained in this case.

### DIFF
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -993,6 +993,19 @@ def test_unsqueeze2():
   verify_torch_and_convert_to_returnn(model_func, inputs=x)
 
 
+def test_transpose_unsqueeze():
+  n_batch, n_time, n_feat = 3, 5, 7
+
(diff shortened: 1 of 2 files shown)


tests/test_layers.py
```diff
@@ -993,6 +993,19 @@ def model_func(wrapped_import, inputs: torch.Tensor):
   verify_torch_and_convert_to_returnn(model_func, inputs=x)
 
 
+def test_transpose_unsqueeze():
```
albertz, line 996: This test is for `Unflatten`? How?

albertz: Ah sorry, I thought of sth else. `Unflatten` is used by `reshape`, and thus here in `unsqueeze`.

albertz (approved):

Issue #84, opened 2022-01-18
issue when indexing merged dim


vieting:
As shown in the test case in #85, there is currently some bug when we index a dim with flattened batch and time.

Code:
```
y = inputs.view(-1, n_in)
idx = torch.arange(n_batch * n_time)
idx = torch.cat([idx] * n_index).reshape(n_batch, n_time * n_index)
x = y[idx.view(-1)]
x = x.view(n_batch, n_time, n_index, n_in)
x = x.permute(2, 0, 1, 3)
out = torch.cat([inputs.unsqueeze(0), x])
```

Log output ([via](https://github.com/rwth-i6/pytorch-to-returnn/runs/4852042902)):
<details>

```
2022-01-18T10:48:42.2941933Z >>> Running with wrapped Torch import, wrapping replacement for PyTorch...
2022-01-18T10:48:42.2942434Z RETURNN input: Data{'data', [B,T|'time:var:extern_data:data'[B],F|F'feature:data'(7)]}
2022-01-18T10:48:42.2943002Z *** root/'Flatten' layer dict: {'class': 'flatten_batch', 'from': 'data', 'axis': 'T', 'batch_major': True}
2022-01-18T10:48:42.2943565Z *** root/'Flatten' FlattenBatchLayer output: [B&Packed{'time:var:extern_data:data'},F|F'feature:data'(7)]
2022-01-18T10:48:42.2944131Z *** root/'Range' layer dict: {'class': 'range', 'limit': 15, 'start': 0, 'delta': 1, 'dtype': None, 'sparse': False}
2022-01-18T10:48:42.2944574Z *** root/'Range' RangeLayer output: [F|'Range:range'(15)]
2022-01-18T10:48:42.2945006Z *** root/'Cat' layer dict: {'class': 'concat', 'from': [('Range', 'F'), ('Range', 'F')]}
2022-01-18T10:48:42.2945425Z *** root/'Cat' ConcatLayer output: [F|'2*Range:range'(30)]
2022-01-18T10:48:42.2945889Z *** root/'Unflatten' layer dict: {'class': 'split_dims', 'from': 'Cat', 'axis': 'F', 'dims': [3, 10]}
2022-01-18T10:48:42.2946418Z *** root/'Unflatten' SplitDimsLayer output: ['Unflatten_split_dims0'(3),F|'Unflatten_split_dims1'(10)]
[...]

PR #85, opened 2022-01-18
indexing merged dim

vieting:
Add a test case which demonstrates the issue #84.

### DIFF
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -4,6 +4,7 @@ Dummies...
 """
 
 from __future__ import annotations
+import numpy
 from typing import Union, Any, Tuple, List, Optional
 from returnn.tf.util.data import Dim
 from ..naming import Naming
@@ -149,10 +150,8 @@ class SizeValue(int):
     return tensor_entry.torch_axis_from_returnn_axis[returnn_axis]
 
   def as_tensor(self):
-    if self.merged_dims and self.originating_tensor is None:
-      from functools import reduce
-      from operator import mul
-      return reduce(mul, [dim.as_tensor() for dim in self.merged_dims], 1)
+    if self.originating_tensor is None and self.merged_dims:
+      return numpy.prod([d.as_tensor() if d.dim_tag.dimension is None else int(d) for d in self.merged_dims])
     assert self.originating_tensor is not None
     from .nn.modules import Length
     tensor = Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -151,13 +151,16 @@ class SizeValue(int):
 
   def as_tensor(self):
     if self.originating_tensor is None and self.merged_dims:
-      return numpy.prod([d.as_tensor() if d.dim_tag.dimension is None else int(d) for d in self.merged_dims])
-    assert self.originating_tensor is not None
-    from .nn.modules import Length
-    tensor = Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
-    if len(tensor.shape) > 0:
-      from . import max
-      tensor = max(tensor)
+      tensor = numpy.prod([
+        d.as_tensor() if isinstance(d, SizeValue) and d.dim_tag.dimension is None else int(d)
+        for d in self.merged_dims])
+    else:
+      assert self.originating_tensor is not None
+      from .nn.modules import Length
+      tensor = Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
+      if len(tensor.shape) > 0:
+        from . import max
+        tensor = max(tensor)
     tensor.fill_(int(self))
     tensor.is_defined = True
     naming = Naming.get_instance()
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -5,9 +5,11 @@ Dummies...
 
 from __future__ import annotations
 import numpy
-from typing import Union, Any, Tuple, List, Optional
+from typing import TYPE_CHECKING, Union, Any, Tuple, List, Optional
 from returnn.tf.util.data import Dim
 from ..naming import Naming
+if TYPE_CHECKING:
+  from .tensor import Tensor
 
 
 def device(name):
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -155,14 +155,13 @@ class SizeValue(int):
     assert self.originating_tensor is not None
     from .nn.modules import Length
     tensor = Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
-    naming = Naming.get_instance()
-    tensor_entry = naming.tensors[tensor]
-    if tensor_entry.returnn_data.have_batch_axis():
+    if len(tensor.shape) > 0:
       from . import max
       tensor = max(tensor)
-      tensor_entry = naming.tensors[tensor]
     tensor.fill_(int(self))
     tensor.is_defined = True
+    naming = Naming.get_instance()
+    tensor_entry = naming.tensors[tensor]
     tensor_entry.is_const = True
     tensor_entry.is_dim = self.dim_tag
     return tensor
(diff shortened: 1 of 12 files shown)


vieting:
@albertz it would be great if you could have a look at this.

albertz:
One problem seems to be the implementation of `MergeDims` ([here](https://github.com/rwth-i6/pytorch-to-returnn/blob/bf5b8ba050751983e3300e828543e5f6f6cff471/pytorch_to_returnn/torch/nn/modules/shape.py#L28)): 

> In case batch axis is involved, this uses RETURNN FlattenBatchLayer.

This is wrong. Or at least inconsistent to what PyTorch does. Why do we even need this separate logic? The RETURNN `MergeDimsLayer` should also handle the case with the batch axis, I think. If not, we should fix that. But using `FlattenBatchLayer` does sth different.

albertz:
The test case is problematic. You should not use `n_batch`, `n_time` inside the function.

vieting:
>The test case is problematic. You should not use n_batch, n_time inside the function model_func. You should redefine it there like:
n_batch, n_time, n_in = y.shape
This is important that we keep the batch dim properly.

Yes, done.

albertz:
I wonder if the `reshape(n_batch, n_time * n_index)` will be handled correctly, i.e. that it gets back the batch dim. This would be via the `Unflatten` layer.

albertz:
Why are the tests not running now?

albertz:
Btw, you should rebase to resolve the conflicts.

vieting:
> One problem seems to be the implementation of MergeDims (here): [...] This is wrong. Or at least inconsistent to what PyTorch does. Why do we even need this separate logic? The RETURNN MergeDimsLayer should also handle the case with the batch axis, I think. If not, we should fix that. But using FlattenBatchLayer does sth different.

Do you mean just removing that special case from `MergeDims` like what I just commited?

albertz:
> > One problem seems to be the implementation of MergeDims (here): [...] using FlattenBatchLayer [...]
 
> Do you mean just removing that special case from `MergeDims` like what I just commited?

Yes. Not sure if that breaks sth else. Or why we even did that in the first place.

But this is required here. Esp the indices (via `arange(n_b * n_t)`) index into the merged B*T axis. This must be merged and not flattened.

albertz:
> I wonder if the `reshape(n_b, n_t * n_index)` will be handled correctly, i.e. that it gets back the batch dim. This would be via the `Unflatten` module. Maybe we need to extend the logic [here](https://github.com/rwth-i6/pytorch-to-returnn/blob/63bcb497f8be66f50a5d8d34ddffdc6b106065ed/pytorch_to_returnn/torch/nn/modules/shape.py#L228).

Yea, it is not correct:
```
*** root/'Unflatten' layer dict: {'class': 'split_dims', 'from': 'Cat', 'axis': 'F', 'dims': [Batch(3), 10]}
*** root/'Unflatten' SplitDimsLayer output: ['Unflatten_split_dims0'(3),F|'Unflatten_split_dims1'(10)]
```

albertz:
Also this is already wrong:
```
*** root/'Range' layer dict: {'class': 'range', 'limit': 15, 'start': 0, 'delta': 1, 'dtype': None, 'sparse': False}
*** root/'Range' RangeLayer output: [F|'Range:range'(15)]
```

albertz:
For reference, in RETURNN, this works now:
```
"flat": {"class": "flatten_batch", "from": "data"},
"length_flat": {"class": "length", "from": "flat", "axis": batch_dim},
"indices_flat": {"class": "rand_int", "shape": sz, "minval": 0, "maxval": "length_flat", "seed": 42},
"output": {"class": "gather", "from": "flat", "axis": batch_dim, "position": "indices_flat"},
```
(Via https://github.com/rwth-i6/returnn/pull/910)

albertz:
Ok, now that #87 is merged, can you rebase, and check if it maybe works already?
If not, why does it not work yet?

vieting:
I think it does not yet work because we have `idx = torch.arange(n_b * n_t)` and when multiplying `SizeValue`s, the `originating_tensor` is lost.

albertz:
Rebase anyway and let's see. I'm not sure.

vieting:
Oh sorry, I accidentally pushed it to my form before.

albertz:
> If `self.originating_tensor` is `None` in `as_tensor`, we could probably check if `self.merged_dims` contains some dims and if so, compute the output based on them.

Yes, this would be one solution, as a fallback.

Sth like:
```
def as_tensor():
  if self.originating_tensor is None and self.merged_dims:
    return numpy.prod([d.as_tensor() if d.dim_tag.dimension is None else int(d) for d in self.merged_dims])
  assert self.originating_tensor is not None
  ...
```

pytorch_to_returnn/torch/_C.py
```diff
@@ -149,6 +149,10 @@ def originating_tensor_axis(self) -> int:
     return tensor_entry.torch_axis_from_returnn_axis[returnn_axis]
 
   def as_tensor(self):
+    if self.merged_dims and self.originating_tensor is None:
+      from functools import reduce
+      from operator import mul
+      return reduce(mul, [dim.as_tensor() for dim in self.merged_dims], 1)
```
albertz, line 155: `as_tensor()` doesn't work (or might not work?) in case of static dims? Not sure.

vieting:
> > If `self.originating_tensor` is `None` in `as_tensor`, we could probably check if `self.merged_dims` contains some dims and if so, compute the output based on them.
> 
> Yes, this would be one solution, as a fallback.
> 
> Sth like:
> 
> ```
> def as_tensor():
>   if self.originating_tensor is None and self.merged_dims:
>     return numpy.prod([d.as_tensor() if d.dim_tag.dimension is None else int(d) for d in self.merged_dims])
>   assert self.originating_tensor is not None
>   ...
> ```

Ok this is probably more accurate than what I pushed.

pytorch_to_returnn/torch/_C.py
albertz: For static dims, we could also return `int(self)`.

vieting:
Now, if we have `as_tensor` of a time dim, we get a batched tensor from `LengthLayer`. I'm not sure what the intended behavior would be in this case as we want to have a scalar. Should it always be the maximum value?

albertz:
> Now, if we have `as_tensor` of a time dim, we get a batched tensor from `LengthLayer`. I'm not sure what the intended behavior would be in this case as we want to have a scalar. Should it always be the maximum value?

The intended behavior is clear. This `as_tensor()` should describe the shape dim, i.e. be a scalar. So then we need some `torch.max`.

pytorch_to_returnn/torch/_C.py
```diff
@@ -149,13 +150,19 @@ def originating_tensor_axis(self) -> int:
...
+    if self.originating_tensor is None and self.merged_dims:
+      return numpy.prod([d.as_tensor() if d.dim_tag.dimension is None else int(d) for d in self.merged_dims])
     assert self.originating_tensor is not None
     from .nn.modules import Length
     tensor = Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
-    tensor.fill_(int(self))
-    tensor.is_defined = True
     naming = Naming.get_instance()
     tensor_entry = naming.tensors[tensor]
+    if tensor_entry.returnn_data.have_batch_axis():
```
albertz, line 160: This is too specific. You want to apply this code always when it is not a scalar.
```suggestion
    if tensor_entry.returnn_data.batch_ndim > 0:
```

albertz: Also, you don't need the `tensor_entry` here. Simply:
```suggestion
    if len(tensor.shape) > 0:
```

pytorch_to_returnn/torch/_C.py
```diff
@@ -149,13 +150,19 @@ def originating_tensor_axis(self) -> int:
...
+      from . import max
+      tensor = max(tensor)
+      tensor_entry = naming.tensors[tensor]
```
albertz, line 163: Move this code before the `tensor_entry`, then you don't need the reassignment here.

albertz:
There is a bug in our `Cat` module. The out shape is wrong.

albertz:
I did some fixes here. I think it's mostly finished, except that I also need to make another fix on RETURNN side in `SplitDimsLayer`.

albertz:
The change in `Unflatten` is causing some problems. I'm working on it.

albertz:
The introduction of dim tags needs some logic similar to returnn-common where we serialize them for the net dict. This is needed for the converter. I will do that later.

albertz:
We also need https://github.com/rwth-i6/returnn/pull/913 here.

PR #87, opened 2022-01-18
SizeValue generic support for dynamic sizes

vieting:
For fixing #84, as demonstrated in #85, we need to save more information in `SizeValue`.

### DIFF
--- a/pytorch_to_returnn/naming/naming.py
+++ b/pytorch_to_returnn/naming/naming.py
@@ -295,7 +295,7 @@ class Naming:
     if x is None:
       return None
     from ..torch._C import SizeValue
-    if isinstance(x, SizeValue):
+    if isinstance(x, SizeValue) and x.dim_tag.dimension is None:
       x = x.as_tensor()
     if isinstance(x, (int, float, bool, numpy.number)):
       return x  # keep as-is for now. would be handled via `get_tensor` later on
--- a/pytorch_to_returnn/naming/naming.py
+++ b/pytorch_to_returnn/naming/naming.py
@@ -325,8 +325,8 @@ class Naming:
       feature_dim_axis=returnn_data.feature_dim_axis_or_unspecified,
       available_for_inference=True)
     entry.returnn_axis_from_torch_axis = {i: i for i in range(returnn_data.batch_ndim)}
-    for dim in range(returnn_data.batch_ndim):
-      tensor.shape[dim].dim_tag = entry.returnn_data.dim_tags[dim]
+    for axis in range(returnn_data.batch_ndim):
+      tensor.shape[axis].dim_tag = entry.returnn_data.dim_tags[axis]
     self.root_namespace.register_input(tensor=entry)
     assert entry.returnn_data
     return entry.returnn_data
--- a/pytorch_to_returnn/naming/naming.py
+++ b/pytorch_to_returnn/naming/naming.py
@@ -325,8 +325,8 @@ class Naming:
       feature_dim_axis=returnn_data.feature_dim_axis_or_unspecified,
       available_for_inference=True)
     entry.returnn_axis_from_torch_axis = {i: i for i in range(returnn_data.batch_ndim)}
-    if entry.returnn_data.have_batch_axis():
-      tensor.shape[entry.returnn_data.batch_dim_axis].is_batch_dim = True
+    for dim in range(returnn_data.batch_ndim):
+      tensor.shape[dim].dim_tag = entry.returnn_data.dim_tags[dim]
     self.root_namespace.register_input(tensor=entry)
     assert entry.returnn_data
     return entry.returnn_data
--- a/pytorch_to_returnn/naming/naming.py
+++ b/pytorch_to_returnn/naming/naming.py
@@ -327,8 +327,6 @@ class Naming:
     entry.returnn_axis_from_torch_axis = {i: i for i in range(returnn_data.batch_ndim)}
     for axis in range(returnn_data.batch_ndim):
       tensor.shape[axis].dim_tag = entry.returnn_data.dim_tags[axis]
-      tensor.shape[axis].tensor = tensor
-      tensor.shape[axis].axis = axis
     self.root_namespace.register_input(tensor=entry)
     assert entry.returnn_data
     return entry.returnn_data
--- a/pytorch_to_returnn/naming/naming.py
+++ b/pytorch_to_returnn/naming/naming.py
@@ -338,7 +338,7 @@ class Naming:
     assert tensor in self.tensors
     entry = self.tensors[tensor]
     assert isinstance(entry, _tensor.TensorEntry)
-    assert not entry.is_param and not entry.is_const and not entry.is_input  # not implemented, although simple...
+    assert not entry.is_param and not entry.is_input  # not implemented, although simple...
     self.outputs.append(entry)
     if self.wrap_to_returnn_enabled:
       self.root_namespace.register_returnn_subnet_output(entry)
--- a/pytorch_to_returnn/naming/naming.py
+++ b/pytorch_to_returnn/naming/naming.py
@@ -285,7 +285,7 @@ class Naming:
     self.const_tensor_cache.append(x)
     return self.tensors[x]
 
-  def _make_tensor(self, x: Union[_types.Tensor, int, float, numpy.number, numpy.ndarray, Any]
+  def _make_tensor(self, x: Union[_types.Tensor, int, float, numpy.number, numpy.ndarray, SizeValue, Any]
                    ) -> Optional[_tensor.TensorEntry]:
     """
     We have to decide which objects we keep as-is (as constant),
@@ -294,6 +294,9 @@ class Naming:
     """
     if x is None:
       return None
+    from ..torch._C import SizeValue
+    if isinstance(x, SizeValue):
+      x = x.as_tensor()
     if isinstance(x, (int, float, bool, numpy.number)):
       return x  # keep as-is for now. would be handled via `get_tensor` later on
     if not self.wrap_to_returnn_enabled:
--- a/pytorch_to_returnn/naming/naming.py
+++ b/pytorch_to_returnn/naming/naming.py
@@ -327,6 +327,8 @@ class Naming:
     entry.returnn_axis_from_torch_axis = {i: i for i in range(returnn_data.batch_ndim)}
     for axis in range(returnn_data.batch_ndim):
       tensor.shape[axis].dim_tag = entry.returnn_data.dim_tags[axis]
+      tensor.shape[axis].tensor = tensor
+      tensor.shape[axis].axis = axis
     self.root_namespace.register_input(tensor=entry)
     assert entry.returnn_data
     return entry.returnn_data
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -148,20 +148,18 @@ class SizeValue(int):
       return int(self) * other
     if type(other) == int and other == 1:
       return self
-    res = SizeValue(super(SizeValue, self).__mul__(other))
-    res.merged_dims = [self, other]
-    res.dim_tag = self.dim_tag * getattr(other, "dim_tag", other)
-    return res
+    merged_dims = [self, other]
+    dim_tag = self.dim_tag * (other.dim_tag if isinstance(other, SizeValue) else other)
+    return SizeValue(super(SizeValue, self).__mul__(other), dim_tag=dim_tag, merged_dims=merged_dims)
 
   def __rmul__(self, other):
     if not isinstance(other, int):  # e.g. a list
       return other * int(self)
     if type(other) == int and other == 1:
       return self
-    res = SizeValue(super(SizeValue, self).__rmul__(other))
-    res.merged_dims = [other, self]
-    res.dim_tag = getattr(other, "dim_tag", other) * self.dim_tag
-    return res
+    merged_dims = [other, self]
+    dim_tag = (other.dim_tag if isinstance(other, SizeValue) else other) * self.dim_tag
+    return SizeValue(super(SizeValue, self).__rmul__(other), dim_tag=dim_tag, merged_dims=merged_dims)
 
 
 def zeros(*shape):
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -162,9 +162,9 @@ class SizeValue(int):
 
   def __repr__(self):
     res = super(SizeValue, self).__repr__()
-    if self.is_batch_dim:
-      res = f"Batch({res})"
-    return res
+    if self.dim_tag is None:
+      return f"?{res}"
+    return f"{self.dim_tag.short_repr()}({res})"
 
   def __mul__(self, other):
     assert isinstance(other, (int, SizeValue)), (  # could be allowed for static dims in the future
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -4,7 +4,8 @@ Dummies...
 """
 
 from __future__ import annotations
-from typing import Union, Any, Tuple
+from typing import Union, Any, Tuple, List, Optional
+from returnn.tf.util.data import Dim
 
 
 def device(name):
@@ -122,15 +123,44 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
 
 class SizeValue(int):
   """
-  We extend this, to store extra information, e.g. such that this reflects the batch-dim.
+  We extend this, to store extra information, e.g. corresponding RETURNN dim tags.
   """
-  is_batch_dim: bool = False
-  merged_dims: list[SizeValue] = []
+  def __new__(cls, x, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None):
+    res = super(SizeValue, cls).__new__(cls, x)
+    res.dim_tag = dim_tag or Dim(dimension=x, description="static_dim")
+    res.merged_dims = merged_dims or []
+    return res
+
+  @property
+  def is_batch_dim(self):
+    if self.dim_tag is None:
+      return False
+    return self.dim_tag.is_batch_dim()
 
   def __repr__(self):
     res = super(SizeValue, self).__repr__()
     if self.is_batch_dim:
       res = f"Batch({res})"
+    return res
+
+  def __mul__(self, other):
+    if not isinstance(other, int):  # e.g. a list
+      return int(self) * other
+    if type(other) == int and other == 1:
+      return self
+    res = SizeValue(super(SizeValue, self).__mul__(other))
+    res.merged_dims = [self, other]
+    res.dim_tag = self.dim_tag * getattr(other, "dim_tag", other)
+    return res
+
+  def __rmul__(self, other):
+    if not isinstance(other, int):  # e.g. a list
+      return other * int(self)
+    if type(other) == int and other == 1:
+      return self
+    res = SizeValue(super(SizeValue, self).__rmul__(other))
+    res.merged_dims = [other, self]
+    res.dim_tag = getattr(other, "dim_tag", other) * self.dim_tag
     return res
 
 
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -144,8 +144,9 @@ class SizeValue(int):
     return res
 
   def __mul__(self, other):
-    if not isinstance(other, int):  # e.g. a list
-      return int(self) * other
+    assert isinstance(other, (int, SizeValue)), (
+      "Multiplying a SizeValue with object of type {} is not allowed because it can lead to bugs, e.g. assumtion of a "
+      "static batch dim.".format(type(other)))
     if type(other) == int and other == 1:
       return self
     merged_dims = [self, other]
@@ -153,8 +154,9 @@ class SizeValue(int):
     return SizeValue(super(SizeValue, self).__mul__(other), dim_tag=dim_tag, merged_dims=merged_dims)
 
   def __rmul__(self, other):
-    if not isinstance(other, int):  # e.g. a list
-      return other * int(self)
+    assert isinstance(other, (int, SizeValue)), (
+      "Multiplying a SizeValue with object of type {} is not allowed because it can lead to bugs, e.g. assumtion of a "
+      "static batch dim.".format(type(other)))
     if type(other) == int and other == 1:
       return self
     merged_dims = [other, self]
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -127,7 +127,7 @@ class SizeValue(int):
   We extend this, to store extra information, e.g. corresponding RETURNN dim tags.
   """
   def __new__(cls, x: int, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None,
-              originating_tensor=None):
+              originating_tensor: Tensor = None):
     res = super(SizeValue, cls).__new__(cls, x)
     res.dim_tag = dim_tag or Dim(dimension=x, description="static_dim")
     res.merged_dims = merged_dims or []
@@ -141,13 +141,13 @@ class SizeValue(int):
     return self.dim_tag.is_batch_dim()
 
   @property
-  def axis(self):
+  def originating_tensor_axis(self) -> int:
     naming = Naming.get_instance()
     return naming.tensors[self.originating_tensor].returnn_data.get_axis_by_tag_name(self.dim_tag.description)
 
   def as_tensor(self):
     from .nn.modules import Length
-    return Length(axis=self.axis).as_returnn_torch_functional()(self.originating_tensor)
+    return Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
 
   def __repr__(self):
     res = super(SizeValue, self).__repr__()
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -151,7 +151,10 @@ class SizeValue(int):
   def as_tensor(self):
     assert self.originating_tensor is not None
     from .nn.modules import Length
-    return Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
+    axis = self.originating_tensor_axis
+    tensor = Length(axis=axis).as_returnn_torch_functional()(self.originating_tensor)
+    tensor.fill_(self.originating_tensor.shape[axis])
+    return tensor
 
   def __repr__(self):
     res = super(SizeValue, self).__repr__()
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -125,10 +125,12 @@ class SizeValue(int):
   """
   We extend this, to store extra information, e.g. corresponding RETURNN dim tags.
   """
-  def __new__(cls, x: int, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None):
+  def __new__(cls, x: int, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None, tensor=None, axis=None):
     res = super(SizeValue, cls).__new__(cls, x)
     res.dim_tag = dim_tag or Dim(dimension=x, description="static_dim")
     res.merged_dims = merged_dims or []
+    res.tensor = tensor
+    res.axis = axis
     return res
 
   @property
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -153,6 +153,10 @@ class SizeValue(int):
     from .nn.modules import Length
     tensor = Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
     tensor.fill_(int(self))
+    tensor.is_defined = True
+    naming = Naming.get_instance()
+    tensor_entry = naming.tensors[self.originating_tensor]
+    tensor_entry.is_const = True
     return tensor
 
   def __repr__(self):
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -142,10 +142,14 @@ class SizeValue(int):
 
   @property
   def originating_tensor_axis(self) -> int:
+    assert self.originating_tensor is not None
     naming = Naming.get_instance()
-    return naming.tensors[self.originating_tensor].returnn_data.get_axis_by_tag_name(self.dim_tag.description)
+    tensor_entry = naming.tensors[self.originating_tensor]
+    returnn_axis = tensor_entry.returnn_data.get_axis_from_description(self.dim_tag)
+    return tensor_entry.torch_axis_from_returnn_axis[returnn_axis]
 
   def as_tensor(self):
+    assert self.originating_tensor is not None
     from .nn.modules import Length
     return Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
 
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -210,3 +210,17 @@ def from_numpy(arr):
   assert isinstance(arr, numpy.ndarray)
   from .tensor import Tensor
   return Tensor(*arr.shape, dtype=str(arr.dtype), numpy_array=arr)
+
+
+def convert_to_tensor(x):
+  import numpy
+  from .tensor import Tensor
+  from .nn.modules.operator import Stack
+  if isinstance(x, Tensor):
+    return x
+  if isinstance(x, (int, float, numpy.number, numpy.ndarray)):
+    return from_numpy(x)
+  if isinstance(x, (list, tuple)):
+    x = [convert_to_tensor(e) for e in x]
+    return Stack(dim=0)(*x)
+  raise TypeError(f"unexpected type {type(x)}")
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -155,8 +155,9 @@ class SizeValue(int):
     tensor.fill_(int(self))
     tensor.is_defined = True
     naming = Naming.get_instance()
-    tensor_entry = naming.tensors[self.originating_tensor]
+    tensor_entry = naming.tensors[tensor]
     tensor_entry.is_const = True
+    tensor_entry.is_dim = self.dim_tag
     return tensor
 
   def __repr__(self):
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -204,6 +204,9 @@ def from_numpy(arr):
     arr = numpy.array(arr, dtype='float32')
   if isinstance(arr, numpy.number):
     arr = numpy.array(arr)
+  if isinstance(arr, (list, tuple)):
+    _entry = from_numpy(arr[0])
+    arr = numpy.array(arr, dtype=_entry.dtype.name)
   assert isinstance(arr, numpy.ndarray)
   from .tensor import Tensor
   return Tensor(*arr.shape, dtype=str(arr.dtype), numpy_array=arr)
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -151,9 +151,8 @@ class SizeValue(int):
   def as_tensor(self):
     assert self.originating_tensor is not None
     from .nn.modules import Length
-    axis = self.originating_tensor_axis
-    tensor = Length(axis=axis).as_returnn_torch_functional()(self.originating_tensor)
-    tensor.fill_(self.originating_tensor.shape[axis])
+    tensor = Length(axis=self.originating_tensor_axis).as_returnn_torch_functional()(self.originating_tensor)
+    tensor.fill_(int(self))
     return tensor
 
   def __repr__(self):
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -125,7 +125,7 @@ class SizeValue(int):
   """
   We extend this, to store extra information, e.g. corresponding RETURNN dim tags.
   """
-  def __new__(cls, x, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None):
+  def __new__(cls, x: int, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None):
     res = super(SizeValue, cls).__new__(cls, x)
     res.dim_tag = dim_tag or Dim(dimension=x, description="static_dim")
     res.merged_dims = merged_dims or []
--- a/pytorch_to_returnn/torch/_C.py
+++ b/pytorch_to_returnn/torch/_C.py
@@ -6,6 +6,7 @@ Dummies...
 from __future__ import annotations
 from typing import Union, Any, Tuple, List, Optional
 from returnn.tf.util.data import Dim
+from ..naming import Naming
 
 
 def device(name):
@@ -125,12 +126,12 @@ class SizeValue(int):
   """
   We extend this, to store extra information, e.g. corresponding RETURNN dim tags.
   """
-  def __new__(cls, x: int, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None, tensor=None, axis=None):
+  def __new__(cls, x: int, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None,
+              originating_tensor=None):
     res = super(SizeValue, cls).__new__(cls, x)
     res.dim_tag = dim_tag or Dim(dimension=x, description="static_dim")
     res.merged_dims = merged_dims or []
-    res.tensor = tensor
-    res.axis = axis
+    res.originating_tensor = originating_tensor
     return res
 
   @property
@@ -139,6 +140,15 @@ class SizeValue(int):
       return False
     return self.dim_tag.is_batch_dim()
 
+  @property
+  def axis(self):
+    naming = Naming.get_instance()
+    return naming.tensors[self.originating_tensor].returnn_data.get_axis_by_tag_name(self.dim_tag.description)
+
+  def as_tensor(self):
+    from .nn.modules import Length
+    return Length(axis=self.axis).as_returnn_torch_functional()(self.originating_tensor)
+
   def __repr__(self):
     res = super(SizeValue, self).__repr__()
     if self.is_batch_dim:
@@ -146,7 +156,7 @@ class SizeValue(int):
     return res
 
   def __mul__(self, other):
-    assert isinstance(other, (int, SizeValue)), (
+    assert isinstance(other, (int, SizeValue)), (  # could be allowed for static dims in the future
       "Multiplying a SizeValue with object of type {} is not allowed because it can lead to bugs, e.g. assumtion of a "
       "static batch dim.".format(type(other)))
     if type(other) == int and other == 1:
@@ -156,7 +166,7 @@ class SizeValue(int):
     return SizeValue(super(SizeValue, self).__mul__(other), dim_tag=dim_tag, merged_dims=merged_dims)
 
   def __rmul__(self, other):
-    assert isinstance(other, (int, SizeValue)), (
+    assert isinstance(other, (int, SizeValue)), (  # could be allowed for static dims in the future
       "Multiplying a SizeValue with object of type {} is not allowed because it can lead to bugs, e.g. assumtion of a "
       "static batch dim.".format(type(other)))
     if type(other) == int and other == 1:
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -45,11 +45,11 @@ def full(size, fill_value, *, out=None, dtype=None, layout=None, device=None, re
   if dtype is None:
     if isinstance(fill_value, int) or isinstance(fill_value, numpy.integer):
       dtype = "int64"
-    elif isinstance(fill_value, (bool, numpy.bool, complex, numpy.complex)):
-      raise NotImplementedError
+    elif isinstance(fill_value, (bool, numpy.bool)):
+      dtype = "bool"
     else:
       dtype = torch.get_default_dtype()
-  out = zeros(*size, dtype=dtype) + fill_value
+  out = zeros(*size, dtype=dtype) + cast(fill_value, dtype)
   out.fill_(fill_value)
   return out
 
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -142,18 +142,11 @@ def is_tensor(obj) -> bool:
 
 
 def sum(input: Tensor, dim: Optional[int] = None, dtype: Optional[Union[str, _dtype]] = None) -> Tensor:
-  assert dim is not None, "not implemented yet"
-  return modules.Reduce(mode="sum", axes=dim)(input)
+  return modules.Reduce(mode="sum", axes=dim).as_returnn_torch_functional()(input)
 
 
 def max(input: Tensor, dim: Optional[int] = None, dtype: Optional[Union[str, _dtype]] = None) -> Tensor:
-  out = input
-  if dim is None:
-    for i in range(len(input.shape)):
-      out = modules.Reduce(mode="max", axes=0)(out)
-  else:
-    out = modules.Reduce(mode="max", axes=dim)(out)
-  return out
+  return modules.Reduce(mode="max", axes=dim).as_returnn_torch_functional()(input)
 
 
 def add(x: Tensor, y: Tensor) -> Tensor:
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -38,6 +38,8 @@ def ones(*size, out=None, dtype=None, layout=None, device=None, requires_grad=Fa
 
 
 def full(size, fill_value, *, out=None, dtype=None, layout=None, device=None, requires_grad=False) -> Tensor:
+  if dtype is None:
+    dtype = type(fill_value).__name__ + "32"
   return zeros(*size, dtype=dtype) + fill_value
 
 
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -139,6 +139,16 @@ def sum(input: Tensor, dim: Optional[int] = None, dtype: Optional[Union[str, _dt
   return modules.Reduce(mode="sum", axes=dim)(input)
 
 
+def max(input: Tensor, dim: Optional[int] = None, dtype: Optional[Union[str, _dtype]] = None) -> Tensor:
+  out = input
+  if dim is None:
+    for i in range(len(input.shape)):
+      out = modules.Reduce(mode="max", axes=0)(out)
+  else:
+    out = modules.Reduce(mode="max", axes=dim)(out)
+  return out
+
+
 def add(x: Tensor, y: Tensor) -> Tensor:
   dtype = result_type(x, y)
   return modules.BinaryOperator(kind="add")(cast(x, dtype), cast(y, dtype))
@@ -351,10 +361,6 @@ def chunk(input: Tensor, chunks: int, dim: int = 0) -> List[Tensor]:
 
 def pad(input: Tensor, pad, mode='constant', value=0) -> Tensor:
   return modules.GenericPadNd(padding=pad, mode=mode, value=value).as_returnn_torch_functional()(input)
-
-
-def max(*inputs: Tensor) -> Tensor:
-  return modules.Max()(*inputs)
 
 
 def linear(input: Tensor, weight: Tensor, bias: Optional[Tensor] = None):
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -188,9 +188,24 @@ def truediv(x: Tensor, y: Tensor) -> Tensor:
   return modules.BinaryOperator(kind="truediv")(cast(x, dtype), cast(y, dtype))
 
 
+def greater(x: Tensor, y: Tensor) -> Tensor:
+  dtype = result_type(x, y)
+  return modules.ComparisonOperator(kind="greater")(cast(x, dtype), cast(y, dtype))
+
+
 def greater_equal(x: Tensor, y: Tensor) -> Tensor:
   dtype = result_type(x, y)
   return modules.ComparisonOperator(kind="greater_equal")(cast(x, dtype), cast(y, dtype))
+
+
+def less(x: Tensor, y: Tensor) -> Tensor:
+  dtype = result_type(x, y)
+  return modules.ComparisonOperator(kind="less")(cast(x, dtype), cast(y, dtype))
+
+
+def less_equal(x: Tensor, y: Tensor) -> Tensor:
+  dtype = result_type(x, y)
+  return modules.ComparisonOperator(kind="less_equal")(cast(x, dtype), cast(y, dtype))
 
 
 def flatten(input: Tensor, start_dim=0, end_dim=-1) -> Tensor:
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -194,22 +194,22 @@ def truediv(x: Tensor, y: Tensor) -> Tensor:
 
 def greater(x: Tensor, y: Tensor) -> Tensor:
   dtype = result_type(x, y)
-  return modules.ComparisonOperator(kind="greater")(cast(x, dtype), cast(y, dtype))
+  return modules.ComparisonOperator(kind="greater").as_returnn_torch_functional()(cast(x, dtype), cast(y, dtype))
 
 
 def greater_equal(x: Tensor, y: Tensor) -> Tensor:
   dtype = result_type(x, y)
-  return modules.ComparisonOperator(kind="greater_equal")(cast(x, dtype), cast(y, dtype))
+  return modules.ComparisonOperator(kind="greater_equal").as_returnn_torch_functional()(cast(x, dtype), cast(y, dtype))
 
 
 def less(x: Tensor, y: Tensor) -> Tensor:
   dtype = result_type(x, y)
-  return modules.ComparisonOperator(kind="less")(cast(x, dtype), cast(y, dtype))
+  return modules.ComparisonOperator(kind="less").as_returnn_torch_functional()(cast(x, dtype), cast(y, dtype))
 
 
 def less_equal(x: Tensor, y: Tensor) -> Tensor:
   dtype = result_type(x, y)
-  return modules.ComparisonOperator(kind="less_equal")(cast(x, dtype), cast(y, dtype))
+  return modules.ComparisonOperator(kind="less_equal").as_returnn_torch_functional()(cast(x, dtype), cast(y, dtype))
 
 
 def flatten(input: Tensor, start_dim=0, end_dim=-1) -> Tensor:
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -29,6 +29,9 @@ def get_default_dtype():
 
 
 def zeros(*size, out=None, dtype=None, layout=None, device=None, requires_grad=False):
+  if size and isinstance(size[0], (tuple, list)):
+    assert len(size) == 1
+    size = tuple(size[0])
   if all(isinstance(x, int) for x in size):
     return Tensor(*size, dtype=dtype)
   if not dtype:
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -78,7 +78,7 @@ def arange(*args, out: Optional[Tensor]=None, dtype: Optional[_dtype]=None, devi
     end, = args
     start = 0
     step = 1
-  mod = modules.Range().as_returnn_torch_functional()
+  mod = modules.Range()
   return mod(end, start, step, dtype, False)
 
 
--- a/pytorch_to_returnn/torch/nn/functional.py
+++ b/pytorch_to_returnn/torch/nn/functional.py
@@ -29,7 +29,11 @@ def get_default_dtype():
 
 
 def zeros(*size, out=None, dtype=None, layout=None, device=None, requires_grad=False):
-  return Tensor(*size, dtype=dtype)
+  if all(isinstance(x, int) for x in size):
+    return Tensor(*size, dtype=dtype)
+  if not dtype:
+    dtype = _default_float_type
+  return modules.FullStatic(fill_value=0, dtype=dtype)(size=size)
 
 
 def ones(*size, out=None, dtype=None, layout=None, device=None, requires_grad=False):
@@ -38,7 +42,9 @@ def ones(*size, out=None, dtype=None, layout=None, device=None, requires_grad=Fa
     size = tuple(size[0])
   if not dtype:
     dtype = _default_float_type
-  return tensor(numpy.ones(size, dtype=dtype))
+  if all(isinstance(x, int) for x in size):
+    return tensor(numpy.ones(size, dtype=dtype))
+  return modules.FullStatic(fill_value=1, dtype=dtype)(size=size)
 
 
 def full(size, fill_value, *, out=None, dtype=None, layout=None, device=None, requires_grad=False) -> Tensor:
@@ -49,6 +55,8 @@ def full(size, fill_value, *, out=None, dtype=None, layout=None, device=None, re
       dtype = "bool"
     else:
       dtype = torch.get_default_dtype()
+  if isinstance(fill_value, (int, float, numpy.number)):
+    return modules.FullStatic(fill_value=fill_value, dtype=dtype)(size=size)
   zeros_ = zeros(*size, dtype=dtype)
   fill_value_ = cast(fill_value, dtype)
   out = zeros_ + fill_value_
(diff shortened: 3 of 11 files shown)


vieting:
I added the dim tags now. The corresponding layer is still missing, we need that as well, right?

pytorch_to_returnn/naming/naming.py
```diff
@@ -325,8 +325,8 @@ def register_input(self, tensor: _types.Tensor, returnn_data: Data) -> Data:
       feature_dim_axis=returnn_data.feature_dim_axis_or_unspecified,
       available_for_inference=True)
     entry.returnn_axis_from_torch_axis = {i: i for i in range(returnn_data.batch_ndim)}
-    if entry.returnn_data.have_batch_axis():
-      tensor.shape[entry.returnn_data.batch_dim_axis].is_batch_dim = True
+    for dim in range(returnn_data.batch_ndim):
```
albertz, line 328: Better call it `axis`, for consistency.
```suggestion
    for axis in range(returnn_data.batch_ndim):
```

pytorch_to_returnn/naming/naming.py
```diff
@@ -325,8 +325,8 @@ def register_input(self, tensor: _types.Tensor, returnn_data: Data) -> Data:
...
+      tensor.shape[dim].dim_tag = entry.returnn_data.dim_tags[dim]
```
albertz, line 329:
```suggestion
      tensor.shape[axis].dim_tag = entry.returnn_data.dim_tags[axis]
```

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +123,46 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
 
 class SizeValue(int):
   """
-  We extend this, to store extra information, e.g. such that this reflects the batch-dim.
+  We extend this, to store extra information, e.g. corresponding RETURNN dim tags.
   """
-  is_batch_dim: bool = False
-  merged_dims: list[SizeValue] = []
+  def __new__(cls, x, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None):
```
albertz, line 128: Add a type to `x`.
```suggestion
  def __new__(cls, x: int, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None):
```

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +123,46 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
+    return self.dim_tag.is_batch_dim()
 
   def __repr__(self):
     res = super(SizeValue, self).__repr__()
     if self.is_batch_dim:
       res = f"Batch({res})"
     return res
 
+  def __mul__(self, other):
+    if not isinstance(other, int):  # e.g. a list
```
albertz, line 147: Do you really get this case? I wonder if we should do sth more clever, or maybe forbid it, because likely whatever this is used for, it would be wrong because it assumes a static batch dim.

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +123,46 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
+      return int(self) * other
+    if type(other) == int and other == 1:
+      return self
+    res = SizeValue(super(SizeValue, self).__mul__(other))
```
albertz, line 151: Better directly pass `merged_dims` and `dim_tag` here.

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +123,46 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
+    res.merged_dims = [self, other]
+    res.dim_tag = self.dim_tag * getattr(other, "dim_tag", other)
```
albertz, line 153: Don't use `getattr`. Check `isinstance(other, SizeValue)` and then directly access the attrib, or otherwise use `other`. So `other.dim_tag if isinstance(other, SizeValue) else other`.

albertz:
> I added the dim tags now. The corresponding layer is still missing, we need that as well, right?

What layer do you mean?

Or you mean the changes to `reshape` or more specifically the `Unflatten` module?

pytorch_to_returnn/torch/_C.py
vieting: Only the "opposite" case in `__rmul__` occurred in the tests ([here](https://github.com/rwth-i6/pytorch-to-returnn/blob/c0b77753d908a3a5a125662da575012b6ae10cba/pytorch_to_returnn/torch/nn/utils/rnn.py#L43)). I added it in `__mul__` for consistency.

albertz: We should maybe replace the `lengths = [time_dim] * batch_dim` by `fill` or some other construct. Because this is wrong.

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +123,44 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
 
 class SizeValue(int):
   """
-  We extend this, to store extra information, e.g. such that this reflects the batch-dim.
+  We extend this, to store extra information, e.g. corresponding RETURNN dim tags.
   """
-  is_batch_dim: bool = False
-  merged_dims: list[SizeValue] = []
```
albertz, line 128: Maybe you anyway want to have this annotation?

vieting:
>>I added the dim tags now. The corresponding layer is still missing, we need that as well, right?

>What layer do you mean?

I mean do we somehow need to keep track of the dim tag's origin?

pytorch_to_returnn/torch/_C.py
vieting: For `is_batch_dim`, this seems outdated. For `merged_dims`, it's already available in the definition of `__new__`. But if you want we can keep it.

pytorch_to_returnn/torch/_C.py
vieting: Yes, that's true. Should we just forbid this case and open an issue for the usage in `pack_sequence`? However, then the test case currently fails.

albertz:
> > > I added the dim tags now. The corresponding layer is still missing, we need that as well, right?
> > 
> > What layer do you mean?
> 
> I mean do we somehow need to keep track of the dim tag's origin?

From our discussion, I remember two use cases:

- `torch.arange` over such `SizeValue`. For that, we can use for example `RangeInAxisLayer` or `RangeFromLengthLayer` + `LengthLayer`. In both cases, despite the dim tag, we also need the original tensor ref (to get some layer, to get the dynamic size from).
- `torch.randint` with `SizeValue` as argument. For that, again `LengthLayer` + `RandIntLayer`. For the `LengthLayer`, we again need the tensor ref.

But basically any other PyTorch function you could potentially call would also need to do that. It should be correct to always use `LengthLayer` basically. I wonder if there is a simple way to have this automatically everywhere. Maybe we should derive `SizeValue` from `Tensor` instead of `int`. But I'm not sure if this would cause other problems.

In any case, all the RETURNN layers already exist, and also all the needed functionality on RETURNN side. This only needs extensions here in pytorch-to-returnn.

pytorch_to_returnn/torch/_C.py
albertz: Does PyCharm correctly finds the definition of `merged_dims`? I know this is the case for `__init__` but I don't know about `__new__`. If so, then this is ok.

pytorch_to_returnn/torch/_C.py
albertz: Just directly fix `pack_sequence` in this PR as well. This should be a very simple fix.

pytorch_to_returnn/torch/_C.py
vieting: Yes it does, so I'll leave it like this.

pytorch_to_returnn/torch/_C.py
vieting: What exactly do you mean by `fill(length, [batch_dim])`? Something like this?
```
lengths.zeros(batch_dim)
lengths.fill_(time_dim)
```

albertz: I meant what `tf.fill` is in TF. It looks like this is `torch.full`. I.e. `torch.full([batch_dim], time_dim)`. That will give you a tensor of shape [batch_dim] with entries time_dim everywhere. Exactly like `[time_dim] * batch_dim` but as a tensor.

vieting:
I started adding `arange` with dynamic limit. Could you please comment on that? Not sure if it's going in the right direction...

pytorch_to_returnn/naming/naming.py
```diff
@@ -325,8 +325,10 @@ def register_input(self, tensor: _types.Tensor, returnn_data: Data) -> Data:
       feature_dim_axis=returnn_data.feature_dim_axis_or_unspecified,
       available_for_inference=True)
     entry.returnn_axis_from_torch_axis = {i: i for i in range(returnn_data.batch_ndim)}
-    if entry.returnn_data.have_batch_axis():
-      tensor.shape[entry.returnn_data.batch_dim_axis].is_batch_dim = True
+    for axis in range(returnn_data.batch_ndim):
+      tensor.shape[axis].dim_tag = entry.returnn_data.dim_tags[axis]
+      tensor.shape[axis].tensor = tensor
+      tensor.shape[axis].axis = axis
```
albertz, line 331: The `axis` should not be needed.

pytorch_to_returnn/naming/naming.py
```diff
@@ -325,8 +325,10 @@ def register_input(self, tensor: _types.Tensor, returnn_data: Data) -> Data:
       feature_dim_axis=returnn_data.feature_dim_axis_or_unspecified,
       available_for_inference=True)
     entry.returnn_axis_from_torch_axis = {i: i for i in range(returnn_data.batch_ndim)}
-    if entry.returnn_data.have_batch_axis():
-      tensor.shape[entry.returnn_data.batch_dim_axis].is_batch_dim = True
+    for axis in range(returnn_data.batch_ndim):
+      tensor.shape[axis].dim_tag = entry.returnn_data.dim_tags[axis]
+      tensor.shape[axis].tensor = tensor
```
albertz, line 330: We can also pass the `tensor` directly in `SizeValue.__new__` (in `Tensor.__init__`), right?

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +123,48 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
+  def __new__(cls, x: int, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None, tensor=None, axis=None):
```
albertz, line 128: `tensor` should be typed.
`axis` should not be needed.

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +123,48 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
   def __repr__(self):
     res = super(SizeValue, self).__repr__()
     if self.is_batch_dim:
       res = f"Batch({res})"
     return res
 
+  def __mul__(self, other):
+    assert isinstance(other, (int, SizeValue)), (
+      "Multiplying a SizeValue with object of type {} is not allowed because it can lead to bugs, e.g. assumtion of a "
+      "static batch dim.".format(type(other)))
```
albertz, line 151: Note that we might want to allow this for static dims. But not so important now.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +24,17 @@ class Range(Module):
   is_original_torch_module = False
 
   def create_returnn_layer_dict(self, limit, start, delta, dtype, sparse=False) -> Dict[str, Any]:
-    return {"class": "range", "limit": limit, "start": start, "delta": delta, "dtype": dtype, "sparse": sparse}
+    if isinstance(limit, SizeValue):
+      limit = Length(axis=limit.axis)(limit.tensor)
```
albertz, line 28: You should never pass `axis` as integer.

albertz: Ah sorry, I was confused. This is here on PyTorch level.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -394,6 +406,22 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
     return {"class": "reduce", "mode": self.mode, "axes": self.axes, "from": self._get_input_layer_name(input)}
 
 
+class Length(Module):
+  """
+  Wraps RETURNN LengthLayer.
+  """
+  is_original_torch_module = False
+
+  def __init__(self, axis: Union[str, Dim]):
```
albertz, line 415: Why `str`?

But maybe we want to allow `Dim` here (a dim tag). Although this mixes PyTorch and RETURNN things but passing `Dim` would only be used internally.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -394,6 +406,22 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
...
+    super(Length, self).__init__()
+    self.axis = axis
+
+  def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
+    return {
+      "class": "length", "axis": self._get_input_axis_to_returnn(input, self.axis),
```
albertz, line 421: When we also have `axis` as `Dim`, then:
```suggestion
      "class": "length", "axis": axis if isinstance(axis, Dim) else self._get_input_axis_to_returnn(input, self.axis),
```

albertz: Note that `_get_input_axis_to_returnn` expects an `int` (PyTorch axis), nothing else, not a `str` and also not a `Dim`.

pytorch_to_returnn/torch/nn/modules/operator.py
albertz: `int` is missing, if this is ever supposed to be used for normal PyTorch code.

pytorch_to_returnn/torch/nn/modules/operator.py
albertz: But anyway. It depends a bit on how we design `Length`. If this can get a `Dim`, we can simply have here:
```suggestion
      limit = Length(axis=limit.dim_tag)(limit.tensor)
```

albertz: As we might need this code maybe in multiple places, and also as we maybe will rewrite this later, I would make a helper function here:
```suggestion
      limit = _size_value_to_tensor(limit)
```
Or maybe as a method of `SizeValue`:
```suggestion
      limit = limit.as_tensor()
```

pytorch_to_returnn/torch/_C.py
albertz: I think we also should rename `tensor` to `originating_tensor` or so.

vieting: For typing of `tensor` we would need to import `Tensor` which leads to a circular import, right?

pytorch_to_returnn/torch/_C.py
vieting: I'll add a comment.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: When implementing it as a method of `SizeValue`, I need to import `Length` in `_C.py` which also leads to a circular import. I could define `_size_value_to_tensor` in this file (or possibly there is a better place?).

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: I used `str` and `Dim` as that's what is allowed in RETURNN. Added `int` now.

pytorch_to_returnn/torch/_C.py
albertz: No, it doesn't need to be imported. You should never import sth just for typing, or maybe only conditionally with `if typing.TYPE_CHECKING:`. You probably should have `from __future__ import annotations` if Python/PyCharm complains.

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +124,57 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
 
 class SizeValue(int):
   """
-  We extend this, to store extra information, e.g. such that this reflects the batch-dim.
+  We extend this, to store extra information, e.g. corresponding RETURNN dim tags.
   """
-  is_batch_dim: bool = False
-  merged_dims: list[SizeValue] = []
+  def __new__(cls, x: int, dim_tag: Optional[Dim] = None, merged_dims: Optional[List[SizeValue]] = None,
+              originating_tensor=None):
```
albertz, line 130: Add type info to `originating_tensor`.

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +124,57 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
+    return res
+
+  @property
+  def is_batch_dim(self):
+    if self.dim_tag is None:
+      return False
+    return self.dim_tag.is_batch_dim()
+
+  @property
+  def axis(self):
```
albertz, line 144: Better also rename, and also add type info:
```suggestion
  def originating_tensor_axis(self) -> int:
```

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: `str` should also work, right? Maybe we would like to get batch shape so we have `axis="B"`?

pytorch_to_returnn/torch/_C.py
vieting: See above, we cannot import `Tensor` here due to a circular import. Or how should we do it?

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +124,57 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
+    naming = Naming.get_instance()
+    return naming.tensors[self.originating_tensor].returnn_data.get_axis_by_tag_name(self.dim_tag.description)
```
albertz, line 146: Do not use `get_axis_by_tag_name`, and also never use `Dim.description`.
Instead, use `get_axis_from_description` (the method name is maybe confusing with `Dim.description` but this is not meant here; this accepts a generic axis description, or directly a dim tag).
Also, you return the RETURNN axis here, which is wrong! You should return the PyTorch axis.
So:
```suggestion
    tensor_entry = naming.tensors[self.originating_tensor]
    returnn_axis = tensor_entry.returnn_data.get_axis_from_description(self.dim_tag)
    return tensor_entry.torch_axis_from_returnn_axis[returnn_axis]
```

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -394,6 +406,24 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
     return {"class": "reduce", "mode": self.mode, "axes": self.axes, "from": self._get_input_layer_name(input)}
 
 
+class Length(Module):
+  """
+  Wraps RETURNN LengthLayer.
+  """
+  is_original_torch_module = False
+
+  def __init__(self, axis: Union[str, Dim, int]):
```
albertz, line 415: Now that you implemented `originating_tensor_axis`, I think to keep it consistent, we should not mix RETURNN stuff here in the PyTorch API, and only allow `int` here.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +24,17 @@ class Range(Module):
   is_original_torch_module = False
 
   def create_returnn_layer_dict(self, limit, start, delta, dtype, sparse=False) -> Dict[str, Any]:
-    return {"class": "range", "limit": limit, "start": start, "delta": delta, "dtype": dtype, "sparse": sparse}
+    if isinstance(limit, SizeValue):
+      limit = limit.as_tensor()
+      return {"class": "range_from_length", "from": self._get_input_layer_name(limit)}
+    else:
```
albertz, line 30: You might want to add some `assert isinstance(limit, int)` here or so.

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +124,57 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
+    return res
+
+  @property
+  def is_batch_dim(self):
+    if self.dim_tag is None:
+      return False
+    return self.dim_tag.is_batch_dim()
+
+  @property
+  def originating_tensor_axis(self) -> int:
```
albertz, line 144: We maybe should add some `assert self.originating_tensor is not None` here.

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +124,57 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
+    naming = Naming.get_instance()
+    return naming.tensors[self.originating_tensor].returnn_data.get_axis_by_tag_name(self.dim_tag.description)
+
+  def as_tensor(self):
```
albertz, line 148: We might want to add some `assert self.originating_tensor is not None` here.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +24,17 @@ class Range(Module):
...
+      return {"class": "range", "limit": limit, "start": start, "delta": delta, "dtype": dtype, "sparse": sparse}
+
+  def _get_output_shape_from_returnn(self, inputs_flat: List, layer: LayerBase
```
albertz, line 33: I assume `List[Tensor]`?

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +24,17 @@ class Range(Module):
...
+                                          ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
```
albertz, line 34: The indentation is off.

pytorch_to_returnn/torch/nn/modules/operator.py
albertz: No, `_get_input_axis_to_returnn` always expects an `int`.
But anyway, see my follow-up comment. I think we always should make `axis` an `int`.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -394,6 +406,24 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
...
+  def __init__(self, axis: Union[str, Dim, int]):
+    super(Length, self).__init__()
+    self.axis = axis
+
+  def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
+    axis = self.axis
+    if isinstance(axis, int):
+      axis = self._get_input_axis_to_returnn(input, axis)
+    return {
+      "class": "length", "axis": axis, "from": self._get_input_layer_name(input)}
```
albertz, line 424: With `axis` always being an `int`, this becomes:
```suggestion
    return {
      "class": "length", "from": self._get_input_layer_name(input),
      "axis": self._get_input_axis_to_returnn(input, self.axis)}
```

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: Actually no. Right now, we get whatever the input args were in `arange`, so a `SizeValue` and `int`s. This is a problem, right? The crash in the test case where the `feed_dict` is not prepared correctly is related to this, I think.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: We allow only `int` now, see comments below.

pytorch_to_returnn/torch/_C.py
albertz: This change for `pack_sequence` is still needed, right?

albertz:
`test_arange` fails:
```
  File "/home/runner/work/pytorch-to-returnn/pytorch-to-returnn/tests/test_layers.py", line 1232, in test_arange.<locals>.model_func
    line: return inputs + torch.reshape(arange, (1, -1))
    locals:
      inputs = <local> <Tensor name:? tensor:(Batch(1),5) returnn_data:'data' [B,F|F'feature:data'(5)] axes id>
      torch = <local> <module 'pytorch_to_returnn.torch' from '/home/runner/work/pytorch-to-returnn/pytorch-to-returnn/pytorch_to_returnn/torch/__init__.py'>
      torch.reshape = <local> <function reshape at 0x7f6230fed670>
      arange = <local> <Tensor name:? tensor:(5,) returnn_data:'Range_output' [T|'Range_input_len'[]] axes id>
...
  File "/home/runner/work/pytorch-to-returnn/pytorch-to-returnn/pytorch_to_returnn/torch/nn/modules/operator.py", line 98, in BinaryOperator.create_returnn_layer_dict
    line: inputs = _unify_tensor_axes_returnn_meta(*inputs)
    locals:
      inputs = <local> (<Tensor name:? tensor:(Batch(1),5) returnn_data:'data' [B,F|F'feature:data'(5)] axes id>, <Tensor name:? tensor:(1,5) returnn_data:'Cast_output' [F|'Unflatten_split_dims0'(1),T|'Range_input_len'[]] axes id>)
      _unify_tensor_axes_returnn_meta = <global> <function _unify_tensor_axes_returnn_meta at 0x7f6238203940>
  File "/home/runner/work/pytorch-to-returnn/pytorch-to-returnn/pytorch_to_returnn/torch/nn/modules/operator.py", line 469, in _unify_tensor_axes_returnn_meta
    line: assert dim_tag.dimension == prev_dim_tag.dimension, f"not matching Data {x.returnn_data}"
    locals:
      dim_tag = <local> Dim{'Range_input_len'[]}
      dim_tag.dimension = <local> None
      prev_dim_tag = <local> Dim{F'feature:data'(5)}
      prev_dim_tag.dimension = <local> 5
AssertionError: not matching Data Data{'Cast_output', [F|'Unflatten_split_dims0'(1),T|'Range_input_len'[]]}
```

This looks not so much related to `Range` but to the broadcasting logic with `_unify_tensor_axes_returnn_meta`. Not sure if we want to fix this here in this PR as well or in another separate PR. A simpler test case for this would probably look like:
```python
inputs  # shape [B,F]
_, n_in = inputs.shape
b = torch.zeros([n_in]).reshape((1, -1))  # [1,F]
return inputs + b  # should broadcast b
```
I think this would fail in the same way.

pytorch_to_returnn/torch/_C.py
vieting: Yes, it is.

vieting:
>inputs  # shape [B,F]
_, n_in = inputs.shape
b = torch.zeros([n_in]).reshape((1, -1))  # [1,F]
return inputs + b  # should broadcast b
I think this would fail in the same way.

For me locally, this passes.

albertz:
> > inputs  # shape [B,F]
> > _, n_in = inputs.shape
> > b = torch.zeros([n_in]).reshape((1, -1))  # [1,F]
> > return inputs + b  # should broadcast b
> > I think this would fail in the same way.
> 
> For me locally, this passes.

Ah yes, there is also another problem:
```
<Tensor name:? tensor:(1,5) returnn_data:'Cast_output' [F|'Unflatten_split_dims0'(1),T|'Range_input_len'[]] axes id>
```
I think the dim 1 in axis 0 is maybe already handled fine. In `_unify_tensor_axes_returnn_meta`, the docstring also says:

>  It also makes sure that broadcasting in the batch-dim axis works correctly
>  by potentially removing such a batch broadcast dim.

But `T|'Range_input_len'[]` looks wrong. This is marked as a dynamic dim. But it should be static (right)?

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +24,19 @@ class Range(Module):
   is_original_torch_module = False
 
   def create_returnn_layer_dict(self, limit, start, delta, dtype, sparse=False) -> Dict[str, Any]:
-    return {"class": "range", "limit": limit, "start": start, "delta": delta, "dtype": dtype, "sparse": sparse}
+    if isinstance(limit, SizeValue):
```
albertz, line 27: I think this only should be done for dynamic dims. Specifically:
```suggestion
    if isinstance(limit, SizeValue) and limit.dim_tag.dimension is None:
```

vieting:
Would you modify the test case here? `test_arange_dyn` fails with
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: You must feed a value for placeholder tensor 'extern_data/placeholders/data/data' with dtype float and shape [?,5]
```

I think this is because for `Range` in `make_output_tensor_from_returnn()`, `input_entries` is all `None` because there are no tensors in it. We might need to pass `limit.as_tensor()` to `Range.create_returnn_layer_dict()`, but then I'm not sure how to infer the correct torch_shape in `Range._get_output_shape_from_returnn()`.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: Yes, this fixes `test_arange` and the issue mentioned in your comment above

>But T|'Range_input_len'[] looks wrong. This is marked as a dynamic dim. But it should be static (right)?

albertz:
> Would you modify the test case here? `test_arange_dyn` fails with
> 
> ```
> tensorflow.python.framework.errors_impl.InvalidArgumentError: You must feed a value for placeholder tensor 'extern_data/placeholders/data/data' with dtype float and shape [?,5]
> ```
> 
> I think this is because for `Range` in `make_output_tensor_from_returnn()`, `input_entries` is all `None` because there are no tensors in it. We might need to pass `limit.as_tensor()` to `Range.create_returnn_layer_dict()`, but then I'm not sure how to infer the correct torch_shape in `Range._get_output_shape_from_returnn()`.

Yes right, this is a problem.

I'm not really sure where to do the `limit.as_tensor()`. The test should not be modified. Corresponding PyTorch code works like that, so we should make sure that it will also work in our converter. This is for the test case and also for Fairseq.

Maybe we can extend the `Naming` logic for `SizeValue`. E.g. `Naming._make_tensor`.

vieting:
>Maybe we can extend the Naming logic for SizeValue. E.g. Naming._make_tensor.

So a `SizeValue` should be returned as is instead of getting `None` in `Naming._make_tensor`?

albertz:
> > Maybe we can extend the Naming logic for SizeValue. E.g. Naming._make_tensor.
> 
> So a `SizeValue` should be returned as is instead of getting `None` in `Naming._make_tensor`? Or should we directly involve `Length` there?

Most other code would not expect `SizeValue`. I think it is easier if this directly uses `size.as_tensor()` and then returns a `TensorEntry`.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -37,6 +37,10 @@ def ones(*size, out=None, dtype=None, layout=None, device=None, requires_grad=Fa
   return tensor(numpy.ones(size, dtype=dtype))
 
 
+def full(size, fill_value, *, out=None, dtype=None, layout=None, device=None, requires_grad=False) -> Tensor:
+  return Tensor(*size, dtype=dtype) + fill_value
```
albertz, line 41: Maybe better using `zeros`.
```suggestion
  return zeros(*size, dtype=dtype) + fill_value
```

vieting:
> >I think this is because for Range in make_output_tensor_from_returnn(), input_entries is all None because there are no tensors in it. We might need to pass limit.as_tensor() to Range.create_returnn_layer_dict(), but then I'm not sure how to infer the correct torch_shape in Range._get_output_shape_from_returnn().
> 
> Yes right, this is a problem.
> 
> I'm not really sure where to do the limit.as_tensor(). The test should not be modified. Corresponding PyTorch code works like that, so we should make sure that it will also work in our converter. This is for the test case and also for Fairseq.
> 
> Maybe we can extend the Naming logic for SizeValue. E.g. Naming._make_tensor.

Btw. what I meant was we could pass `limit.as_tensor()` to the call of `Range()` from `functional.arange()`. So this would not change anything for the PyTorch Code.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +24,19 @@ class Range(Module):
   is_original_torch_module = False
 
   def create_returnn_layer_dict(self, limit, start, delta, dtype, sparse=False) -> Dict[str, Any]:
-    return {"class": "range", "limit": limit, "start": start, "delta": delta, "dtype": dtype, "sparse": sparse}
+    if isinstance(limit, SizeValue) and limit.dim_tag.dimension is None:
```
albertz, line 27: When we change `Naming._make_tensor` as discussed, we would not get `SizeValue` here but instead we would get a `Tensor`.

albertz:
> Btw. what I meant was we could pass `limit.as_tensor()` to the call of `Range()` from `functional.arange()`. So this would not change anything for the PyTorch Code.

Ok, this would fix `arange`. But what if you call some module directly and pass a `SizeValue`? This might happen.

vieting:
> Most other code would not expect SizeValue. I think it is easier if this directly uses size.as_tensor() and then returns a TensorEntry.

I get a circular import again here, if I want to import `SizeValue` in `naming.py` to do the instance check.

vieting:
> Ok, this would fix arange. But what if you call some module directly and pass a SizeValue? This might happen.

`Range` is not an original PyTorch module, so I don't think this can happen.

albertz:
> 



> > Ok, this would fix arange. But what if you call some module directly and pass a SizeValue? This might happen.
> 
> `Range` is not an original PyTorch module, so I don't think this can happen.

Not for `Range` but for any other PyTorch module.

albertz:
> > Most other code would not expect SizeValue. I think it is easier if this directly uses size.as_tensor() and then returns a TensorEntry.
> 
> I get a circular import again here, if I want to import `SizeValue` in `naming.py` to do the instance check.

So, what's the problem? Why don't you just fix it?

You anyway should avoid to import too many things in the top level of a module. I prefer to only import things locally inside a function when it is not really used anywhere else. And if only used for type annotations, then also no generic import (as I commented before).

albertz:
> I don't know how check whether it's static and how to infer the PyTorch shape in this case...

In what case? You mean in `Range`? It should be simple: When it is `isinstance(limit, Tensor)`, then it is dynamic, otherwise it is `isinstance(limit, int)` (and never `isinstance(limit, SizeValue)`).

vieting:
> >I don't know how check whether it's static and how to infer the PyTorch shape in this case...
>
> In what case? You mean in Range? It should be simple: When it is isinstance(limit, Tensor), then it is dynamic, otherwise it is isinstance(limit, int) (and never isinstance(limit, SizeValue)).

Yes, for `Range`. But no, also if it's a static dim, we end up with a tensor for `limit`. Like in `test_arange`.

Also, I don't know how to set `torch_shape` correctly in `_get_output_shape_from_returnn`.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +23,18 @@ class Range(Module):
...
+    if isinstance(limit, Tensor) and limit.dim_tag.dimension is None:
+      return {"class": "range_from_length", "from": self._get_input_layer_name(limit)}
+    else:
+      assert isinstance(limit, int)
+      return {"class": "range", "limit": limit, "start": start, "delta": delta, "dtype": dtype, "sparse": sparse}
+
+  def _get_output_shape_from_returnn(self,
+                                    inputs_flat: List, layer: LayerBase
+                                    ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    torch_shape = ((int(inputs_flat[0]) - int(inputs_flat[1])) // inputs_flat[2],)
```
vieting, line 35: And here, how to infer the integer values for the torch side if we get a `Tensor`.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +23,18 @@ class Range(Module):
   is_original_torch_module = False
 
   def create_returnn_layer_dict(self, limit, start, delta, dtype, sparse=False) -> Dict[str, Any]:
-    return {"class": "range", "limit": limit, "start": start, "delta": delta, "dtype": dtype, "sparse": sparse}
+    if isinstance(limit, Tensor) and limit.dim_tag.dimension is None:
```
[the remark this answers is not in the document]
albertz: As said, this check should have been done in `Naming._make_tensor` already.

pytorch_to_returnn/naming/naming.py
```diff
@@ -294,6 +294,9 @@ def _make_tensor(self, x: Union[_types.Tensor, int, float, numpy.number, numpy.n
     """
     if x is None:
       return None
+    from ..torch._C import SizeValue
+    if isinstance(x, SizeValue):
```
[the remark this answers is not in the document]
vieting: Ah yes, this should solve the first problem regarding the check.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +23,18 @@ class Range(Module):
...
+      return {"class": "range_from_length", "from": self._get_input_layer_name(limit)}
+    else:
+      assert isinstance(limit, int)
+      return {"class": "range", "limit": limit, "start": start, "delta": delta, "dtype": dtype, "sparse": sparse}
+
+  def _get_output_shape_from_returnn(self,
+                                    inputs_flat: List, layer: LayerBase
```
albertz, line 33: `List` should specify the type.

albertz: Also, the indentation is wrong.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: This here remains also after fixing `_make_tensor`

albertz: Simply `inputs_flat[0].numpy()`?
Btw, you better should name the args:
```python
limit, start, delta, *_ = inputs_flat
```

albertz: Ah then we need to fix the `_numpy_buffer` for this tensor in `SizeValue.as_tensor()`.

vieting: I pushed this anyway now, just for reference. The test passes, but I think this is not correct yet. The output also seems wrong
```
layer root/'output' output: Data{'output_output', [T|'Range_input_len'[]], dtype='int32'}
```

vieting: > Ah then we need to fix the _numpy_buffer for this tensor in SizeValue.as_tensor().

Ah sorry, this did not show up before. I'll fix this.

vieting: Yes, now the `torch_shape` is correct.

But the output seems still incorrect, doesn't it? I mean why is it `T`?
```
layer root/'output' output: Data{'output_output', [T|'Range_input_len'[]], dtype='int32'}
```

albertz: `T` is just the marker for the `time_dim_axis`. This should not be relevant here.

vieting: Alright 👍

pytorch_to_returnn/torch/_C.py
```diff
@@ -122,17 +124,64 @@ class Size(tuple):  # type: Tuple[SizeValue, ...]
...
+    tensor_entry = naming.tensors[self.originating_tensor]
+    returnn_axis = tensor_entry.returnn_data.get_axis_from_description(self.dim_tag)
+    return tensor_entry.torch_axis_from_returnn_axis[returnn_axis]
+
+  def as_tensor(self):
+    assert self.originating_tensor is not None
+    from .nn.modules import Length
+    axis = self.originating_tensor_axis
+    tensor = Length(axis=axis).as_returnn_torch_functional()(self.originating_tensor)
+    tensor.fill_(self.originating_tensor.shape[axis])
```
albertz, line 156: Or simply:
```suggestion
    tensor.fill_(int(self))
```

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -394,6 +409,22 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
...
+  is_original_torch_module = False
+
+  def __init__(self, axis: int):
+    super(Length, self).__init__()
+    self.axis = axis
+
+  def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
+    return {
+      "class": "length", "axis": self._get_input_axis_to_returnn(input, self.axis),
+      "from": self._get_input_layer_name(input)}
```
albertz, line 425: We need to somehow specify `int32` for the returned tensor dtype.

albertz: Hm or maybe not.

albertz: No, actually, I think the tensor in the end will be created via:
```
res = module.make_output_tensor_from_returnn(inputs_flat=inputs_flat, layer=layer)
```
And that will do:
```
tensor = Tensor(*torch_shape, numpy_array=numpy_array, dtype=layer.output.dtype)
```

vieting: On RETURNN side or torch side?

vieting:
`Tensor.__init__` uses `self._numpy_buffer = numpy.zeros(shape, dtype=dtype) if numpy_array is None else numpy_array`. I think this is wrong if we have dynamic dims, right? Is there a RETURNN layer which can create zeros of a given shape?

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -37,6 +37,10 @@ def ones(*size, out=None, dtype=None, layout=None, device=None, requires_grad=Fa
   return tensor(numpy.ones(size, dtype=dtype))
 
 
+def full(size, fill_value, *, out=None, dtype=None, layout=None, device=None, requires_grad=False) -> Tensor:
+  return zeros(*size, dtype=dtype) + fill_value
```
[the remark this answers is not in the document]
vieting: I guess only in the case that `dtype=None`, right?

albertz:
> `Tensor.__init__` uses `self._numpy_buffer = numpy.zeros(shape, dtype=dtype) if numpy_array is None else numpy_array`. I think this is wrong if we have dynamic dims, right? Is there a RETURNN layer which can create zeros of a given shape?

Why is this wrong? `_numpy_buffer` is about the static Torch buffer, nothing dynamic.

vieting:
> Why is this wrong? _numpy_buffer is about the static Torch buffer, nothing dynamic.

Ah no sorry, but `Constant.create_returnn_layer_dict` needs to be fixed such that we don't just have a `ConstantLayer` with th numpy zeros array printed as `value`, right?

albertz:
> Ah no sorry, but `Constant.create_returnn_layer_dict` needs to be fixed such that we don't just have a `ConstantLayer` with th numpy zeros array printed as `value`, right?

How exactly do we get `Constant` here? Also, I don't exactly understand the problem.

vieting:
> Why is value not a Numpy array here?

`tensor.numpy()` returns a numpy array (`array([0., 0., 0.], dtype=float32)`), but `batch_axis=0`. And after the `if batch_axis is not None` branch, `value=0`.

vieting:
Should `zeros` maybe directly call `Constant`? And then we could have a more generic layer definition there, where not just the numpy array is printed in the net dict but it could make use of the `SizeValue`s passed for the shape.

albertz:
> > Why is value not a Numpy array here?
> 
> `tensor.numpy()` returns a numpy array (`array([0., 0., 0.], dtype=float32)`), but `batch_axis=0`. And after the `if batch_axis is not None` branch, `value=0`.

Ah, we probably need this there:
```
value = numpy.array(value[0])  # remove batch axis
```

> Should `zeros` maybe directly call `Constant`? And then we could have a more generic layer definition there, where not just the numpy array is printed in the net dict but it could make use of the `SizeValue`s passed for the shape.
> 
> Or maybe a different module because that could not use the `ConstantLayer` I guess.

Our current `Constant` module does not really work this way. It expects an existing `Tensor` and this is assumed to be constant and then wrapped via the module. Also, the current logic in `prepare_tensor_as_input` and `Constant` preserves the original tensor. I think some other code might depend on this behavior. I think I would try to avoid changing this now because this might cause again some other unrelated problems.

The `ConstantLayer` on RETURNN side now has become more generic. It accepts a generic `shape` option where you could pass dynamic dim tags. But this requires that the `value` is the same in the whole tensor. But this should always be the case when this is used currently.

So we might change the current implementation of our `Constant` module to also handle more custom `tensor` inputs. But I think this is not needed at the moment for the cases we have here.

vieting:
> So we might change the current implementation of our Constant module to also handle more custom tensor inputs. But I think this is not needed at the moment for the cases we have here.

This is because it's the batch dim here, right? For other dynamic dims, that would not work, would it? But alright, if it's fine here, then this could be done later if needed.

albertz:
> > So we might change the current implementation of our Constant module to also handle more custom tensor inputs. But I think this is not needed at the moment for the cases we have here.
> 
> This is because it's the batch dim here, right? For other dynamic dims, that would not work, would it? But alright, if it's fine here, then this could be done later if needed.

Right. Yes, for other dynamic dims, this does not work yet. But this would be simple to add via the `shape` option. We can do that later once we need it. Or now if you want.

albertz:
The recent changes unveil further bugs now:
```
  File "/home/runner/work/pytorch-to-returnn/pytorch-to-returnn/pytorch_to_returnn/torch/nn/utils/rnn.py", line 12, in pack_padded_sequence
    line: for frame in range(lengths[0]):
    locals:
      frame = <not found>
      range = <builtin> <class 'range'>
      lengths = <local> <Tensor name:? tensor:(Batch(3),) returnn_data:'add_output' [B] axes id>
TypeError: 'Tensor' object cannot be interpreted as an integer
```
Such `for` loop cannot work on dynamic tensors. `pack_padded_sequence` has to be reimplemented.

pytorch_to_returnn/torch/nn/functional.py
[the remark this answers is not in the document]
vieting: I think that's not done in the original PyTorch code
```
>>> torch.full((5,), np.int32(42)).dtype                                                                                                                                                                                             
torch.int64
```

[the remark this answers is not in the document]
albertz: Oh I see. But it's still wrong then. In case `fill_value` is an `int`, your current code would return a float tensor.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -38,6 +38,8 @@ def ones(*size, out=None, dtype=None, layout=None, device=None, requires_grad=Fa
 
 
 def full(size, fill_value, *, out=None, dtype=None, layout=None, device=None, requires_grad=False) -> Tensor:
+  if dtype is None:
+    dtype = type(fill_value).__name__ + "32"
```
albertz, line 42: This is too hacky and wrong in a couple of cases (e.g. when `fill_value` is a `numpy.int32`).

albertz: Also, it still does not follow the PyTorch logic.

vieting: Yes, it also seemed a bit hacky to me. But we to just take `type(fill_value)` we would need to extend `_dtype` in `_C.py` and if we don't take care of the 32 this will also not work.

albertz: The check should be:
```suggestion
    if isinstance(fill_value, int) or isinstance(fill_value, numpy.integer):
      dtype = "int64"
    else:
      dtype = "float32"  # actually torch.get_default_dtype()
```

albertz: Do you already have `get_default_dtype`? Maybe add it and just return "float32" there for now.

The logic here might still be incomplete, e.g. for "bool".

pytorch_to_returnn/torch/nn/utils/rnn.py
```diff
@@ -1,15 +1,31 @@
...
 from ..modules.rnn import PackedSequence
 from ..modules.shape import FlattenBatch
 from ...tensor import Tensor
 from ..._C import from_numpy
+from .... import torch
 from ....naming import Naming
 
 
+def _batch_sizes_from_lengths(lengths: Tensor) -> Tensor:
+  indices = torch.arange(torch.max(lengths))  # [T]
```
vieting, line 12: I think this does not work properly. `Range._get_output_shape_from_returnn` gets a `Tensor` for `limit`, but `limit.numpy()=0` because the `numpy_buffer` it not set correctly after `Reduce`.

pytorch_to_returnn/torch/nn/functional.py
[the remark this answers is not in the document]
vieting: I mark this as resolved because there are follow up threads.

pytorch_to_returnn/torch/nn/utils/rnn.py
albertz: But then this sounds like yet another bug. This should work. Maybe you can make a separate test case for this.

vieting: I pushed something to set the numpy buffer which fixes this here

pytorch_to_returnn/torch/nn/utils/rnn.py
```diff
@@ -1,15 +1,31 @@
...
+  indices_bc = indices.unsqueeze(0)  # [1,T]
+  l_bc = lengths.unsqueeze(1)  # [B,1]
+  mask = indices_bc < l_bc  # [B,T]
```
vieting, line 15: The broadcasting here seems not to work yet.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -49,7 +49,9 @@ def full(size, fill_value, *, out=None, dtype=None, layout=None, device=None, re
       raise NotImplementedError
     else:
       dtype = torch.get_default_dtype()
-  return zeros(*size, dtype=dtype) + fill_value
+  out = zeros(*size, dtype=dtype) + fill_value
+  out.fill_(fill_value)
```
albertz, line 53: This should not be done. This assumes a kind of strange behavior of `fill_`. This also should be done automatically already in case `fill_value` is constant.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -152,6 +154,7 @@ def max(input: Tensor, dim: Optional[int] = None, dtype: Optional[Union[str, _dt
       out = modules.Reduce(mode="max", axes=0)(out)
   else:
     out = modules.Reduce(mode="max", axes=dim)(out)
+  out.fill_(input._numpy_buffer.max(dim))
```
albertz, line 157: Same here. Also, you should not access `_numpy_buffer`.

pytorch_to_returnn/torch/nn/functional.py
vieting: What else do you suggest? This is not done automatically yet apparently.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -37,6 +41,19 @@ def ones(*size, out=None, dtype=None, layout=None, device=None, requires_grad=Fa
   return tensor(numpy.ones(size, dtype=dtype))
 
 
+def full(size, fill_value, *, out=None, dtype=None, layout=None, device=None, requires_grad=False) -> Tensor:
+  if dtype is None:
+    if isinstance(fill_value, int) or isinstance(fill_value, numpy.integer):
+      dtype = "int64"
+    elif isinstance(fill_value, (bool, numpy.bool, complex, numpy.complex)):
+      raise NotImplementedError
```
albertz, line 49: Just make:
```suggestion
      dtype = "bool"
```

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -37,6 +41,19 @@ def ones(*size, out=None, dtype=None, layout=None, device=None, requires_grad=Fa
...
+    else:
+      dtype = torch.get_default_dtype()
+  out = zeros(*size, dtype=dtype) + fill_value
```
albertz, line 52:
```suggestion
  out = zeros(*size, dtype=dtype) + torch.cast(fill_value, dtype)
```

pytorch_to_returnn/torch/nn/functional.py
albertz: But then we should fix this. Why is this not done automatically? This is wrong when this is a constant.

pytorch_to_returnn/torch/nn/utils/rnn.py
```diff
@@ -1,15 +1,31 @@
...
+  batch_sizes = torch.sum(mask.int(), dim=0)  # [T]
+  return batch_sizes
+
+
 def pack_padded_sequence(input: Tensor, lengths, batch_first=False, enforce_sorted=True) -> PackedSequence:
-  batch_sizes = []
-  for frame in range(lengths[0]):
-    batch_sizes.append(sum([x > frame for x in lengths]))
+  batch_sizes = None
+  if isinstance(lengths, Tensor):
```
albertz, line 22: We should not need this here. `_batch_sizes_from_lengths` should work for both cases.

pytorch_to_returnn/torch/nn/utils/rnn.py
albertz: Maybe you can also make a separate test case specifically for sth like `torch.arange(torch.max(lengths))`.

pytorch_to_returnn/torch/nn/utils/rnn.py
albertz: If there is some issue specifically about broadcasting (not related to previous other problems), then you should make a separate test case only for the broadcasting, and also fix that.

albertz: Maybe you mean this from the failing test case:
```
  File "/home/runner/work/pytorch-to-returnn/pytorch-to-returnn/pytorch_to_returnn/torch/nn/utils/rnn.py", line 15, in _batch_sizes_from_lengths
    line: mask = indices_bc < l_bc  # [B,T]
    locals:
      mask = <not found>
      indices_bc = <local> <Tensor name:? tensor:(1,5) returnn_data:'Unflatten_output' [F|'Unflatten_split_dims0'(1),T|'Range_input_len'[]] axes id>
      l_bc = <local> <Tensor name:? tensor:(Batch(3),1) returnn_data:'Unflatten_1_output' [B,F|'Unflatten_1_split_dims1'(1)] axes id>
```
In the following back trace, there is some problem in `_unify_tensor_axes_returnn_meta`. It creates a `ReinterpretDataLayer` with `'set_axes': {'F': 'B'}`, which does not make sense (and that is also the error then).
We should make a separate test case just reproducing the problem here.

albertz: In general, `_unify_tensor_axes_returnn_meta` should not reassign any axis ("F" here) to the batch axis, or neither the other way around (reassign the batch axis ("B") to some other axis). Maybe if you just exclude this (reassignments on/to the batch axis), it's already fine.

pytorch_to_returnn/torch/nn/functional.py
albertz: In `Module.make_output_tensor_from_returnn` we do (for constant inputs). And this is the right place.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -22,7 +23,21 @@ class Range(Module):
...
+    else:
+      assert isinstance(limit, int)
+      return {"class": "range", "limit": limit, "start": start, "delta": delta, "dtype": dtype, "sparse": sparse}
+
+  def _get_output_shape_from_returnn(self,
+                                     inputs_flat: List[Optional[Union[Tensor, int, bool]]], layer: LayerBase
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    limit, start, delta, *_ = inputs_flat
+    if isinstance(limit, Tensor):
+      limit = limit.numpy()
```
albertz, line 37: This might need some extra error check to catch an error earlier on that this value is known. Maybe we need to introduce a new flag `Tensor.is_defined` or so. If this is not known, this could be due to a bug (your current case).

But in principle, when it (`limit`) is really dynamic, the question is what we should do then. I don't see a good solution. Maybe we would just not allow this.

pytorch_to_returnn/torch/nn/utils/rnn.py
vieting: I added this test case (`test_arange_from_lengths`), it currently works.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: Yes not allowing it for now seems reasonable to me. The PR is already diverging significantly from what we originally started it for.

pytorch_to_returnn/torch/nn/utils/rnn.py
albertz: You say the test `test_arange_from_lengths` works? But so what's different here then? Maybe `lengths` is not constant. Where does this fail, in what test? Where does `lengths` come from?

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: How would this flag `Tensor.is_defined` be handled? Every time there is a module which does not take care of the numpy buffer it would be set to `False` and subsequently remain so?

pytorch_to_returnn/torch/nn/utils/rnn.py
vieting: Oh sorry, I still have my numpy array modifications active which you wanted to remove.

pytorch_to_returnn/torch/nn/modules/operator.py
albertz: And here in this code, you would just make:
```python
assert limit.is_defined
```

pytorch_to_returnn/torch/nn/utils/rnn.py
vieting: Hm no I didn't

vieting: But this just does not fail because the torch shape is not checked. That's wrong `(0,)` here as well. In the other test case `test_packed_sequence_2` subsequently something fails because of the wrong torch shape I think.

albertz: Then add some `assert arange.shape == ...` to the test `test_arange_from_lengths`.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: And the `False` would be set in `Module.make_output_tensor_from_returnn`?

pytorch_to_returnn/torch/nn/utils/rnn.py
vieting: Yes, this makes the problem obvious. Just pushed it.

pytorch_to_returnn/torch/nn/modules/operator.py
albertz: `is_defined = numpy_buffer is not None` sets both True and False.

albertz:
Btw, in `Constant.make_output_tensor_from_returnn`, we do:
```python
entry.is_const = False  # this depends on the (dynamic) batch dim
```
I wonder if this causes maybe some of the problems here. I wonder what happens if we just remove this.

pytorch_to_returnn/torch/nn/functional.py
vieting: So the input to `Length` is the original input tensor which is not marked as `is_const`. This is why this is not done automatically.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -130,6 +146,16 @@ def sum(input: Tensor, dim: Optional[int] = None, dtype: Optional[Union[str, _dt
   return modules.Reduce(mode="sum", axes=dim)(input)
 
 
+def max(input: Tensor, dim: Optional[int] = None, dtype: Optional[Union[str, _dtype]] = None) -> Tensor:
+  out = input
+  if dim is None:
```
albertz, line 151: 1. You should not make this a separate case here but let `Reduce` handle this logic.

2. Btw, in `sum(...)` here, you should remove the `assert dim is not None`. `Reduce` should handle that, or if this is not implemented yet, then `Reduce` should fail.

3. Also, I just noticed that `Reduce` is implemented incorrectly. It should not accept `str` for `axes`. Instead though, it should also accept `None` or a `List[int]`. And it must use `_get_input_axis_to_returnn` in `Reduce.create_returnn_layer_dict` for the axes.

pytorch_to_returnn/torch/nn/functional.py
albertz: But the `zeros` produces a `Constant` which is always marked as `is_const` or not, even when the shape is not const? I think it should.
And the `fill_value` should also marked `is_const`, and then the `cast` as well, and then the `add`.

pytorch_to_returnn/torch/nn/functional.py
vieting: Ok, I fixed this now.

pytorch_to_returnn/torch/nn/functional.py
vieting: The output of `zeros` is not marked as const.

```
>>> zeros_ = zeros(*size, dtype=dtype)
>>> naming.tensors[zeros_]
<TensorEntry name:? tensor:(Batch(3),) returnn_data:None>
>>> naming.tensors[zeros_].is_const
False
```

vieting: This was done from inside `full`

vieting: Also, `fill_value` is via `Length`, so not sure if it should be marked `is_const`.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -395,14 +395,18 @@ class Reduce(Module):
   """
   is_original_torch_module = False
 
-  def __init__(self, mode: str, axes: Union[str, int]):
+  def __init__(self, mode: str, axes: Optional[List[int]]):
```
albertz, line 398: `int` itself would also be allowed here.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -395,14 +395,18 @@ class Reduce(Module):
...
     super(Reduce, self).__init__()
     self.mode = mode
     self.axes = axes
 
   def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
-    return {"class": "reduce", "mode": self.mode, "axes": self.axes, "from": self._get_input_layer_name(input)}
+    if self.axes is None:
+      axes = [self._get_input_axis_to_returnn(input, axis) for axis in range(input.ndim)]
+    else:
+      axes = [self._get_input_axis_to_returnn(input, axis) for axis in self.axes]
```
albertz, line 408:
```suggestion
    axes = self.axes
    if axes is None:
      axes = range(input.ndim)
    if isinstance(axes, int):
      axes = [axes]
    axes = [self._get_input_axis_to_returnn(input, axis) for axis in axes]
```

pytorch_to_returnn/torch/nn/functional.py
vieting: I removed `entry.is_const=False` as in your related comment and set `is_const=True` and `is_defined=True` in `SizeValue.as_tensor()`. Still, this does not work. If I set a breakpoint in `torch.full` when running `test_arange_from_lengths`, for both `zeros_` and `fill_value_` we have `is_const=False` and `is_defined=False`.

albertz: > Also, `fill_value` is via `Length`, so not sure if it should be marked `is_const`.
> set `is_const=True` and `is_defined=True` in `SizeValue.as_tensor()`

Why `is_defined`? Not only that but also the buffer.

Then the `Length` output tensor should not be `is_const=False` anymore, right?

> Still, this does not work. If I set a breakpoint in `torch.full` when running `test_arange_from_lengths`, for both `zeros_` and `fill_value_` we have `is_const=False` and `is_defined=False`.

But where does this get set to `is_const=False`?

albertz:
I might checkout the branch later tonight and try some debugging and maybe follow-up fixes myself. So make sure that you get my commits (pull from the branch).

pytorch_to_returnn/torch/nn/functional.py
vieting: Thanks a lot for your commits! I think they fixed this, so I'm resolving this thread.

pytorch_to_returnn/torch/nn/utils/rnn.py
vieting: I added a test case `test_broadcasting_with_lengths`. However, even if I avoid reassigning the batch axis, the problem persists. Should I anyway push a commit to avoid reassigning the batch axis?

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -194,22 +194,22 @@ def truediv(x: Tensor, y: Tensor) -> Tensor:
 
 def greater(x: Tensor, y: Tensor) -> Tensor:
   dtype = result_type(x, y)
-  return modules.ComparisonOperator(kind="greater")(cast(x, dtype), cast(y, dtype))
+  return modules.ComparisonOperator(kind="greater").as_returnn_torch_functional()(cast(x, dtype), cast(y, dtype))
```
albertz, line 197: Note that `.as_returnn_torch_functional` is never needed when the module itself already has `is_original_torch_module = False` set. I would avoid it in that case as it makes the code a bit harder to read.

This simply marks that the converter logic should not try to search for a corresponding PyTorch module call when comparing the results.

pytorch_to_returnn/torch/nn/utils/rnn.py
albertz: This is fixed now.

albertz (approved):

albertz:
I think this looks all fine now.

Issue #89, opened 2022-01-24
Checks for non-deterministic modules


vieting:
When a given PyTorch model is converted, we have checks at different levels to ensure that the outout of the converted RETURNN model matches the reference. In case of non-deterministic modules, this is obviously bound to fail.

In an offline discussion, we thought about tracking non-deterministic modules and adding their output to the `feed_dict` when the RETURNN model is executed.

vieting:
Fixed with #90.

PR #90, opened 2022-01-24
Checks for non-deterministic modules

vieting:
As described in #89, we currently do not account for non-deterministic modules when checking against the reference.

I would like to add `torch.randint`, therefore this PR should also introduce a logic for non-deterministic modules

### DIFF
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -168,10 +168,10 @@ class Converter:
         visited_entries.add(entry_)
         if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
           non_deterministic_tensor_entries.append(entry_)
+          continue
         for call_ in entry_.output_from_calls:
-          for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
-            if prev_entry_ not in visited_entries:
-              entries_to_visit.append(prev_entry_)
+          for prev_entry_ in [pe for pe in call_.inputs_tensor_deps if pe not in visited_entries]:
+            entries_to_visit.append(prev_entry_)
 
       for entry in non_deterministic_tensor_entries:
         if not all(call_.module.module.is_deterministic for call_ in entry.output_from_calls):
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -144,7 +144,7 @@ class Converter:
     assert self._returnn_net_dict, "Call run() first."
     return self._returnn_net_dict
 
-  def _make_tf_feed_dict(self, input: Data, out_entry=None):
+  def _make_tf_feed_dict(self, input: Data, out_entry: Optional[TensorEntry] = None):
     assert input.batch_ndim == len(self._inputs_np.shape)
     assert all(input.batch_shape[i] in {None, self._inputs_np.shape[i]} for i in range(input.batch_ndim))
     n_batch = 1 if input.batch_dim_axis is None else self._inputs_np.shape[input.batch_dim_axis]
@@ -157,11 +157,8 @@ class Converter:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
     if out_entry:
-      def _output_from_non_deterministic(tensor_entry_: TensorEntry):
-        return not all(call_.module.module.is_deterministic for call_ in tensor_entry_.output_from_calls)
-
       for entry in [out_entry.output_from_calls[0].inputs_args[0]]:  # TODO: this is just the case I'm looking for right now, needs to be generalized
-        if _output_from_non_deterministic(entry):
+        if not all(call_.module.module.is_deterministic for call_ in entry.output_from_calls):
           assert entry.validated_to_torch_tf_feed_dict is not None
           d.update(entry.validated_to_torch_tf_feed_dict)
     return d
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -97,6 +97,7 @@ class Converter:
     self._torch_namespace = None  # type: Optional[Naming]
     self._out_returnn_np = None  # type: Optional[numpy.ndarray]
     self._returnn_net_dict = None  # type: Optional[Dict[str, Dict[str, Any]]]
+    self._non_deterministic_layer_outputs = None  # type: Optional[Dict[str, numpy.ndarray]]
 
   def _transform_inputs_data_kwargs(self,
                                     inputs: numpy.ndarray,
@@ -262,6 +263,7 @@ class Converter:
         torch_mods_with_params = naming.get_modules_with_params_by_abs_name()
         print(">>>> Modules with params:")
         pprint(dict(torch_mods_with_params))
+        self._non_deterministic_layer_outputs = naming.non_deterministic_layer_outputs
 
       feed_dict = self._make_tf_feed_dict(x, out_returnn_)
       y_, y_size = session.run((y.placeholder, y.size_placeholder.as_dict()), feed_dict=feed_dict)
@@ -320,10 +322,9 @@ class Converter:
       x = network.extern_data.get_default_input_data()
       y = network.get_default_output_layer().output
       feed_dict = self._make_tf_feed_dict(x)
-      naming = Naming.get_instance()
       feed_dict.update({
         network.layers[layer_name].output.placeholder: value for layer_name, value in
-        naming.non_deterministic_layer_outputs.items()})
+        self._non_deterministic_layer_outputs.items()})
       y_, y_size = session.run((y.placeholder, y.size_placeholder.as_dict()), feed_dict=feed_dict)
       assert isinstance(y_, numpy.ndarray)
       print("Output shape:", y_.shape)
@@ -365,10 +366,9 @@ class Converter:
       x = network.extern_data.get_default_input_data()
       y = network.get_default_output_layer().output
       feed_dict = self._make_tf_feed_dict(x)
-      naming = Naming.get_instance()
       feed_dict.update({
         network.layers[layer_name].output.placeholder: value for layer_name, value in
-        naming.non_deterministic_layer_outputs.items()})
+        self._non_deterministic_layer_outputs.items()})
       y_, y_size = session.run((y.placeholder, y.size_placeholder.as_dict()), feed_dict=feed_dict)
       assert isinstance(y_, numpy.ndarray)
       print("Output shape:", y_.shape)
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -158,22 +158,21 @@ class Converter:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
     if out_entry:
-      visited_entries = set()
       non_deterministic_tensor_entries =  []
-
-      def _get_non_deterministic_tensor_entries(entry_: TensorEntry):
-        """
-        Go recursively through all tensor entries to find non deterministic ones
-        """
+      entries_to_visit = [out_entry]
+      visited_entries = set()
+      while entries_to_visit:
+        entry_ = entries_to_visit.pop(0)
+        if entry_ in visited_entries:
+          continue
+        visited_entries.add(entry_)
         if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
           non_deterministic_tensor_entries.append(entry_)
         for call_ in entry_.output_from_calls:
           for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
-            if prev_entry_ != entry_ and prev_entry_ not in visited_entries:
-              visited_entries.add(prev_entry_)
-              _get_non_deterministic_tensor_entries(prev_entry_)
+            if prev_entry_ not in visited_entries:
+              entries_to_visit.append(prev_entry_)
 
-      _get_non_deterministic_tensor_entries(out_entry)
       for entry in non_deterministic_tensor_entries:
         if not all(call_.module.module.is_deterministic for call_ in entry.output_from_calls):
           assert entry.validated_to_torch_tf_feed_dict is not None
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -170,8 +170,9 @@ class Converter:
           non_deterministic_tensor_entries.append(entry_)
           continue
         for call_ in entry_.output_from_calls:
-          for prev_entry_ in [pe for pe in call_.inputs_tensor_deps if pe not in visited_entries]:
-            entries_to_visit.append(prev_entry_)
+          for prev_entry_ in call_.inputs_tensor_deps:
+            if prev_entry_ not in visited_entries:
+              entries_to_visit.append(prev_entry_)
 
       for entry in non_deterministic_tensor_entries:
         if not all(call_.module.module.is_deterministic for call_ in entry.output_from_calls):
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -157,7 +157,20 @@ class Converter:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
     if out_entry:
-      for entry in [out_entry.output_from_calls[0].inputs_args[0]]:  # TODO: this is just the case I'm looking for right now, needs to be generalized
+      def _get_non_deterministic_tensor_entries(entry_: TensorEntry):
+        """
+        Go recursively through all tensor entries to find non deterministic ones
+        """
+        if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
+          return [entry_]
+        non_det_mods = []
+        for call_ in entry_.output_from_calls:
+          for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
+            if prev_entry_ != entry_:
+              non_det_mods += _get_non_deterministic_tensor_entries(prev_entry_)
+        return non_det_mods
+
+      for entry in _get_non_deterministic_tensor_entries(out_entry):
         if not all(call_.module.module.is_deterministic for call_ in entry.output_from_calls):
           assert entry.validated_to_torch_tf_feed_dict is not None
           d.update(entry.validated_to_torch_tf_feed_dict)
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -320,6 +320,10 @@ class Converter:
       x = network.extern_data.get_default_input_data()
       y = network.get_default_output_layer().output
       feed_dict = self._make_tf_feed_dict(x)
+      naming = Naming.get_instance()
+      feed_dict.update({
+        network.layers[layer_name].output.placeholder: value for layer_name, value in
+        naming.non_deterministic_layer_outputs.items()})
       y_, y_size = session.run((y.placeholder, y.size_placeholder.as_dict()), feed_dict=feed_dict)
       assert isinstance(y_, numpy.ndarray)
       print("Output shape:", y_.shape)
@@ -361,6 +365,10 @@ class Converter:
       x = network.extern_data.get_default_input_data()
       y = network.get_default_output_layer().output
       feed_dict = self._make_tf_feed_dict(x)
+      naming = Naming.get_instance()
+      feed_dict.update({
+        network.layers[layer_name].output.placeholder: value for layer_name, value in
+        naming.non_deterministic_layer_outputs.items()})
       y_, y_size = session.run((y.placeholder, y.size_placeholder.as_dict()), feed_dict=feed_dict)
       assert isinstance(y_, numpy.ndarray)
       print("Output shape:", y_.shape)
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -158,6 +158,8 @@ class Converter:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
     if out_entry:
+      visited_entries = set()
+
       def _get_non_deterministic_tensor_entries(entry_: TensorEntry):
         """
         Go recursively through all tensor entries to find non deterministic ones
@@ -167,7 +169,8 @@ class Converter:
         non_det_mods = []
         for call_ in entry_.output_from_calls:
           for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
-            if prev_entry_ != entry_:
+            if prev_entry_ != entry_ and prev_entry_ not in visited_entries:
+              visited_entries.add(prev_entry_)
               non_det_mods += _get_non_deterministic_tensor_entries(prev_entry_)
         return non_det_mods
 
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -159,22 +159,22 @@ class Converter:
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
     if out_entry:
       visited_entries = set()
+      non_deterministic_tensor_entries =  []
 
       def _get_non_deterministic_tensor_entries(entry_: TensorEntry):
         """
         Go recursively through all tensor entries to find non deterministic ones
         """
         if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
-          return [entry_]
-        non_det_mods = []
+          non_deterministic_tensor_entries.append(entry_)
         for call_ in entry_.output_from_calls:
           for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
             if prev_entry_ != entry_ and prev_entry_ not in visited_entries:
               visited_entries.add(prev_entry_)
-              non_det_mods += _get_non_deterministic_tensor_entries(prev_entry_)
-        return non_det_mods
+              _get_non_deterministic_tensor_entries(prev_entry_)
 
-      for entry in _get_non_deterministic_tensor_entries(out_entry):
+      _get_non_deterministic_tensor_entries(out_entry)
+      for entry in non_deterministic_tensor_entries:
         if not all(call_.module.module.is_deterministic for call_ in entry.output_from_calls):
           assert entry.validated_to_torch_tf_feed_dict is not None
           d.update(entry.validated_to_torch_tf_feed_dict)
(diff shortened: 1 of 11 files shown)


vieting:
The logic I added in `_make_tf_feed_dict` is similar to `Module.check_call_returnn_outputs_to_prev_torch`. However, `torch_naming=None` in this case, so the subsequent steps are failing.

vieting:
I updated the PR and for the `_run_torch_returnn_drop_in` this works now. The feed dict creation needs to be generalized, but do you think the rest is legit?

Then we need a concept for the RETURNN standalone runs. Should we also feed the tensor there? Then we need to do this via absolute layer names I guess. Or we skip the check in case there is a non-deterministic module to make sure that the RETURNN code also runs standalone and not just with the tensor in the `feed_dict`.

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +156,14 @@ def _make_tf_feed_dict(self, input: Data):
     else:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
+    if out_entry:
+      def _output_from_non_deterministic(tensor_entry_: TensorEntry):
+        return not all(call_.module.module.is_deterministic for call_ in tensor_entry_.output_from_calls)
+
+      for entry in [out_entry.output_from_calls[0].inputs_args[0]]:  # TODO: this is just the case I'm looking for right now, needs to be generalized
```
albertz, line 163: TODO

pytorch_to_returnn/converter/converter.py
```diff
@@ -144,7 +144,7 @@ def returnn_net_dict(self) -> Dict[str, Dict[str, Any]]:
     assert self._returnn_net_dict, "Call run() first."
     return self._returnn_net_dict
 
-  def _make_tf_feed_dict(self, input: Data):
+  def _make_tf_feed_dict(self, input: Data, out_entry=None):
```
albertz, line 147: What is `out_entry`? What type?

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +156,14 @@ def _make_tf_feed_dict(self, input: Data):
     else:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
+    if out_entry:
+      def _output_from_non_deterministic(tensor_entry_: TensorEntry):
```
albertz, line 160: I don't think it makes it really easier to read to make this a function. Esp that this function is only called in one place.

pytorch_to_returnn/import_wrapper/context.py
```diff
@@ -9,7 +9,8 @@ def __init__(self, *,
                wrap_mods_direct: Set[str] = None, wrap_mods_indirect: Set[str] = None,
                wrap_mods_alternatives: Dict[str, str] = None,
                keep_as_is_types: Iterable[type] = (),
-               explicit_wrapped_types: Dict[type, "ExplicitWrappedType"] = None):
+               explicit_wrapped_types: Dict[type, "ExplicitWrappedType"] = None,
+               explicit_wrapped_objects: Dict[Callable, Callable] = None):
```
albertz, line 13:
```suggestion
               explicit_wrapped_objects: Dict[Any, Callable[[Any],Any]] = None):
```

pytorch_to_returnn/import_wrapper/context.py
```diff
@@ -118,12 +121,28 @@ def _raise_not_implemented(*args):
...
+    def __init__(self, func, func_name):
+      super(Functional, self).__init__()
+      self.func = func
+      self.func_name = func_name
+
+    def forward(self, *args, **kwargs):
+      return self.func(*args, **kwargs)
+
+    @classmethod
+    def has_torch_forward(cls) -> bool:
```
albertz, line 134: Why do you need this?

pytorch_to_returnn/import_wrapper/context.py
```diff
@@ -118,12 +121,28 @@ def _raise_not_implemented(*args):
     torch.nn.Module: ExplicitWrappedType(torch.nn.Module, WrappedModuleBase, wrap=_raise_not_implemented),
   }
 
+  class Functional(WrappedModuleBase):
```
albertz, line 124: I think I would move this out somewhere. Maybe to `base_wrappers.py`.

pytorch_to_returnn/import_wrapper/context.py
```diff
@@ -118,12 +121,28 @@ def _raise_not_implemented(*args):
...
+      self.func_name = func_name
+
+    def forward(self, *args, **kwargs):
+      return self.func(*args, **kwargs)
+
+    @classmethod
+    def has_torch_forward(cls) -> bool:
+      return False
+
+  obj_map = {torch.randint: (Functional, (torch.randint, "RandInt"))}
```
albertz, line 137: This is too specific. It should be just a single function, maybe with signature `(obj, *, ctx: WrapCtx, name: str)`.

And the logic you have in `wrap`, you would move here.

pytorch_to_returnn/import_wrapper/wrap.py
```diff
@@ -14,6 +14,11 @@ def wrap(obj, *, name: str, ctx: WrapCtx):
     return obj
   if isinstance(obj, ctx.keep_as_is_types):
     return obj
+  if isinstance(obj, Hashable) and obj in ctx.explicit_wrapped_objects:
+    func, args = ctx.explicit_wrapped_objects[obj]
+    args = [make_wrapped_function(arg, name=name, ctx=ctx) if isinstance(arg, _FuncTypes) else arg for arg in args]
+    wrapped_class = make_wrapped_class(func, name=name, ctx=ctx)
+    obj = wrapped_class(*args)
```
albertz, line 21: This is too specific. `explicit_wrapped_objects` would simply give you a function which does that.
```suggestion
  if isinstance(obj, Hashable) and obj in ctx.explicit_wrapped_objects:
    func = ctx.explicit_wrapped_objects[obj]
    obj = func(obj, name=name, ctx=ctx)
```

pytorch_to_returnn/import_wrapper/context.py
vieting: You mean `base_wrappers/function.py`? Or maybe somewhere in `torch_wrappers/`?

pytorch_to_returnn/import_wrapper/context.py
albertz: Like:
```
def _make_wrapped_func(obj, ctx: WrapCtx, name: str):
  wrapped_func = make_wrapped_function(obj, name=name, ctx=ctx)
  wrapped_class = make_wrapped_class(Functional, name="_.Functional", ctx=ctx)
  return wrapped_class(func=wrapped_func, func_name=name)
```

pytorch_to_returnn/import_wrapper/context.py
albertz: Yea maybe `torch_wrappers/function.py`.

pytorch_to_returnn/converter/converter.py
albertz: Why not `out_entry.output_from_calls` directly?

pytorch_to_returnn/import_wrapper/context.py
```diff
@@ -24,6 +25,7 @@ def __init__(self, *,
       These will not be wrapped using :class:`WrappedModule` but can map to some other namespace.
     :param keep_as_is_types: e.g. {torch.device, torch.dtype, torch.Size}
     :param explicit_wrapped_types: e.g. {torch.Tensor: ExplicitWrappedType(WrappedTorchTensor...)}
+    :param explicit_wrapped_objects: e.g. {torch.randint: lambda func: Functional(func, "randint")}
```
albertz, line 28: The comment should match what I suggested below.

pytorch_to_returnn/naming/call.py
```diff
@@ -160,6 +160,21 @@ def apply_call(self) -> _types.Tensor:
...
+          assert call_idx < len(torch_mod_calls)
+          torch_mod_call = torch_mod_calls[call_idx]
+          assert len(torch_mod_call.orig_outputs_flat) == len(self.outputs_flat)
+          for tensor_returnn, tensor_torch in zip(self.outputs_flat, torch_mod_call.orig_outputs_flat):
+            if tensor_returnn:
+              tensor_returnn.validated_to_torch = True
+              if not tensor_returnn.validated_to_torch_tf_feed_dict:
+                tensor_returnn.validated_to_torch_tf_feed_dict = {}
+              torch_np = tensor_torch.detach().cpu().numpy()
+              tensor_returnn.validated_to_torch_tf_feed_dict[tensor_returnn.returnn_data.placeholder] = torch_np
```
albertz, line 177: And additionally, here you maybe can store this in some dict such that we can use it later for the RETURNN standalone run, like:
```
naming.random_layer_outputs[layer.get_absolute_name()] = torch_np
```

pytorch_to_returnn/import_wrapper/context.py
vieting: Without this, the namespace is flattened according to `_flatten_namespace_for_mod` and the entry is not correctly registered in `make_module_call`.

albertz:
> I updated the PR and for the `_run_torch_returnn_drop_in` this works now. The feed dict creation needs to be generalized, but do you think the rest is legit?

Yes, looks good! Only some smaller comments (see inline comments).

> Then we need a concept for the RETURNN standalone runs. Should we also feed the tensor there? Then we need to do this via absolute layer names I guess. Or we skip the check in case there is a non-deterministic module to make sure that the RETURNN code also runs standalone and not just with the tensor in the `feed_dict`.

I made one suggestion how to keep the reference per absolute layer name. Then putting those into the corresponding feed dict of the RETURNN standalone run should be simple. It might already work.

If this turns out to be complicated or difficult, as you say, we might drop the equal check.

pytorch_to_returnn/import_wrapper/context.py
vieting: Sounds good. Just the name has to be resolved somehow. This way, we cannot `get_module_by_abs_id_name` based on the name `RandInt` from the corresponding module. This has to be hard-coded here somehow, right?

pytorch_to_returnn/import_wrapper/context.py
albertz: I don't exactly understand.

`_flatten_namespace_for_mod` will flatten when `has_torch_forward()` returns False, which is what you do here. So the opposite of what you say?

Also, why does `_flatten_namespace_for_mod` matter here?

But anyway, maybe add such a comment into the code.

pytorch_to_returnn/import_wrapper/context.py
albertz: What is `name` (`func_name`) actually here? Sth like `torch.rand_int`, right?

pytorch_to_returnn/import_wrapper/context.py
vieting: With the change here, `mod.has_torch_forward()` is `False`, so `_flatten_namespace_for_mod` will return `False`.

This matters in `make_module_call`:

```
    if parent_namespace is root_namespace and _flatten_namespace_for_mod(module_entry):
      # Special case, somewhat nicer to flatten the namespace for this case.
      root_namespace.assign_call(entry)
      namespace = root_namespace
    else:
      namespace = parent_namespace.register_sub_call(entry)
```

Only in the second branch, a corresponding entry in the namespace's `childs_by_name` is created which we need lateron.

vieting: Yes, I'll add a comment.

pytorch_to_returnn/import_wrapper/context.py
vieting: I used `"RandInd"` because that's the name of the module which does the RETURNN wrapping and therefore when this module looks for the corresponding entry in the namespace it will be by the name `"RandInt"`.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -46,6 +46,33 @@ def _get_output_shape_from_returnn(self,
     return torch_shape, returnn_axis_from_torch_axis
 
 
+class RandInt(Module):
+  is_original_torch_module = False
+  is_deterministic = False
+
```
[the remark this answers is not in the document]
albertz:
```python
def get_returnn_name(self):
  return "rand_int"
```

pytorch_to_returnn/import_wrapper/context.py
albertz: But what is `name` here?

vieting: It's whatever `name` is in `wrap()`. In this specific case, that is `"pytorch_to_returnn.import_wrapper._torch_traced.torch.randint"`.

pytorch_to_returnn/torch/nn/modules/operator.py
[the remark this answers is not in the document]
vieting: Yes, good point :+1:

pytorch_to_returnn/import_wrapper/context.py
vieting: This can be solved by overwriting `RandInt.get_returnn_name` as you suggested below.

pytorch_to_returnn/converter/converter.py
vieting: The `RandInt` module is not in there, we have to do it recursively.
```
>>> out_entry.output_from_calls                                                                                                                                                                                                 
[<CallEntry 'add' <ModuleEntry <BinaryOperator>> (depth 0)>,
 <CallEntry 'output' <ModuleEntry <Copy>> (depth None)>]
```

I added a commit for this.

vieting:
Apparently, different tests access the same `Naming` instance. Did you know this? Do you know how to avoid it? Currently, the tests after `test_randint` fail because `"randing"` is still in `naming.non_deterministic_layer_outputs` but obviously not in the network anymore.

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +157,24 @@ def _make_tf_feed_dict(self, input: Data):
...
+        """
+        Go recursively through all tensor entries to find non deterministic ones
+        """
+        if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
+          return [entry_]
+        non_det_mods = []
+        for call_ in entry_.output_from_calls:
+          for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
+            if prev_entry_ != entry_:
+              non_det_mods += _get_non_deterministic_tensor_entries(prev_entry_)
```
albertz, line 171: You probably visit many entries multiple times, and thus exponentially many times. To avoid this, you need to keep track a set of visited entries.

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +157,24 @@ def _make_tf_feed_dict(self, input: Data):
     else:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
+    if out_entry:
+      def _get_non_deterministic_tensor_entries(entry_: TensorEntry):
+        """
+        Go recursively through all tensor entries to find non deterministic ones
+        """
+        if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
```
albertz, line 165: I think you could also stop doing the recursive search when you see some entry with `validated_to_torch` because that means that this should already be part of the feed dict.

pytorch_to_returnn/naming/naming.py
```diff
@@ -21,6 +21,7 @@ class Naming:
   modules: OrderedDict[_types.Module, _module.ModuleEntry]
   inputs: List[_tensor.TensorEntry]
   outputs: List[_tensor.TensorEntry]
+  non_deterministic_layer_outputs: Dict[str, numpy.ndarray] = {}
```
albertz, line 24: You should **not** (never) make a mutable object instance (e.g. a dict) as a class attribute. Otherwise multiple `Naming` instances might share the same `non_deterministic_layer_outputs` which is not what you want. This should be assigned in the `__init__`. Only the type annotation can stay here.

albertz:
> Apparently, different tests access the same `Naming` instance. Did you know this? Do you know how to avoid it? Currently, the tests after `test_randint` fail because `"randint"` is still in `naming.non_deterministic_layer_outputs` but obviously not in the network anymore.

Are you sure? Note that for `non_deterministic_layer_outputs` you had exactly this bug (see my comment).

vieting:
> > Apparently, different tests access the same `Naming` instance. Did you know this? Do you know how to avoid it? Currently, the tests after `test_randint` fail because `"randint"` is still in `naming.non_deterministic_layer_outputs` but obviously not in the network anymore.
> 
> Are you sure? Note that for `non_deterministic_layer_outputs` you had exactly this bug (see my comment).

Yes right, fixed this now.

pytorch_to_returnn/converter/converter.py
vieting: I don't think so. E.g. in `test_batch_norm`, the `feed_dict` so far only contained the data and the data size placeholder. However, `entry_.validated_to_torch_tf_feed_dict` for the batch norm's output entry contains the batch norm output placeholder. So this is not part of the feed dict yet and it is deterministic, so we should not skip this part in the `session.run` for the check.

vieting:
To me this seems ready now. If you want I can clean up the history to separate between the handling of non-deterministic modules and adding `torch.randint`, otherwise we just squash everything.

pytorch_to_returnn/converter/converter.py
albertz: But when this is deterministic and we have the output value, then whatever comes before this doesn't matter?

pytorch_to_returnn/naming/naming.py
```diff
@@ -462,10 +464,18 @@ def get_module_by_abs_id_name(self, name: str) -> _types.Module:
...
-          possible_namespaces = [name_ for name_ in potential_namespaces if part_name in name_.childs_by_name]
-          assert possible_namespaces, (
+          potential_namespaces_ = []
+          for name_ in potential_namespaces:
+            if part_name in name_.childs_by_name:
+              potential_namespaces_.append(name_.childs_by_name[part_name])
+            else:  # look for explicitly wrapped objects using :class:`Functional`
+              for child_ in name_.childs_by_name.values():
+                for mod_ in child_.modules:
+                  if part_name in getattr(mod_.module, "func_name", ""):
```
albertz, line 474: I think I would not use `getattr` but instead explicitly check for our `Functional` class.

albertz:
> To me this seems ready now. If you want I can clean up the history to separate between the handling of non-deterministic modules and adding `torch.randint`, otherwise we just squash everything.

Looks mostly good, except some smaller things, see my remaining comments.

I think we can just squash everything then.

pytorch_to_returnn/converter/converter.py
vieting: By "is deterministic", I mean that the module is not random. If we skip whatever comes before, we don't verify that part against the torch ref. So e.g. the batch norm computation would not be part of the check while I think it should be part.

albertz: I don't exactly understand. This here is just for creating the feed dict, not about the check. Also `_make_tf_feed_dict` is just used for the whole network check. And given some output entry, here we search for all needed feed inputs to calculate that output entry. When you feed in some intermediate entry (because it is random), everything which comes before that is not relevant to calculate the output.

vieting: > When you feed in some intermediate entry (because it is random), everything which comes before that is not relevant to calculate the output.

Exactly. And I'm arguing that `entry_.validated_to_torch_tf_feed_dict` also contains tensors which are output of regular, deterministic modules. If we would add all of them, they would not be considered in the check.

albertz: But I did not say to add all. I said the opposite. I would stop the recursive search for this case and not continue adding things.

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +157,27 @@ def _make_tf_feed_dict(self, input: Data):
     else:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
+    if out_entry:
+      visited_entries = set()
+
+      def _get_non_deterministic_tensor_entries(entry_: TensorEntry):
+        """
+        Go recursively through all tensor entries to find non deterministic ones
+        """
```
albertz, line 166: The check on `visited_entries` makes already sense right here:
```python
if entry_ in visited_entries:
  return []
visited_entries.add(entry_)
```

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +157,27 @@ def _make_tf_feed_dict(self, input: Data):
...
+        if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
+          return [entry_]
+        non_det_mods = []
+        for call_ in entry_.output_from_calls:
+          for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
+            if prev_entry_ != entry_ and prev_entry_ not in visited_entries:
+              visited_entries.add(prev_entry_)
```
albertz, line 173: I would move that `visited_entries` logic above (see other comment).
And also, I don't think you need the `prev_entry_ != entry_` check anymore then.

pytorch_to_returnn/converter/converter.py
vieting: Isn't it possible that there is some random module before that?

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +157,27 @@ def _make_tf_feed_dict(self, input: Data):
     else:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
+    if out_entry:
+      visited_entries = set()
```
albertz, line 161: I think I would directly collect the `non_deterministic_tensor_entries` right here in a list, and not recursively returning lists in the function and then doing `non_det_mods += ...`.
So you would add here:
```python
non_deterministic_tensor_entries =  []
```

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +157,27 @@ def _make_tf_feed_dict(self, input: Data):
...
+          return [entry_]
+        non_det_mods = []
+        for call_ in entry_.output_from_calls:
+          for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
+            if prev_entry_ != entry_ and prev_entry_ not in visited_entries:
+              visited_entries.add(prev_entry_)
+              non_det_mods += _get_non_deterministic_tensor_entries(prev_entry_)
+        return non_det_mods
+
+      for entry in _get_non_deterministic_tensor_entries(out_entry):
```
albertz, line 177:
```suggestion
      _get_non_deterministic_tensor_entries(out_entry)

      for entry in non_deterministic_tensor_entries:
```

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +157,27 @@ def _make_tf_feed_dict(self, input: Data):
     else:
       for i, size in input.size_placeholder.items():
         d[size] = [self._inputs_np.shape[input.get_batch_axis(i)]] * n_batch  # not so relevant
+    if out_entry:
+      visited_entries = set()
+
+      def _get_non_deterministic_tensor_entries(entry_: TensorEntry):
```
albertz, line 163: Note, actually I would not introduce such a function and do this recursively at all. Recursion is somewhat inefficient. Rather, I would start defining a queue, like:
```python
entries_to_visit = [out_entry]
```

And then you have a loop:
```python
while entries_to_visit:
  entry_ = entries_to_visit.pop(0)
  ...
```
And then the code from the function follows.
You would replace the recursive call by:
```python
entries_to_visit.append(prev_entry_)
```

pytorch_to_returnn/converter/converter.py
albertz: Why does it matter? `out_entry` depends on this module, so when you feed this, whatever comes before does not matter.

vieting: I thought it would be possible to have something like input -> some random module -> some deterministic module -> output, where `entry_.validated_to_torch_tf_feed_dict` is set for the output of the deterministic module. Then, we would still need to go deeper because we need to add the random module's output to the `feed_dict`. If you say this case cannot occur, I'll add the suggested stop condition.

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +157,26 @@ def _make_tf_feed_dict(self, input: Data):
...
+      visited_entries = set()
+      while entries_to_visit:
+        entry_ = entries_to_visit.pop(0)
+        if entry_ in visited_entries:
+          continue
+        visited_entries.add(entry_)
+        if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
+          non_deterministic_tensor_entries.append(entry_)
+        for call_ in entry_.output_from_calls:
+          for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
```
albertz, line 172: I would not create this temp list here. You can simply move the check `if isinstance(prev_entry_, TensorEntry)` into the loop.

But also, I think you should not loop over `input_args`. That can be arbitrarily nested. But even when you flatten this, it could still miss things (also e.g. from kwargs). The right way would be to iterate over `inputs_tensor_deps`.

pytorch_to_returnn/converter/converter.py
albertz: Ah sorry, my initial comment confused this function with some other case. I thought this is for intermediate checking, not for the whole network checking. For the whole network checking, yes, you are right, nothing else is needed to be added here anymore.

pytorch_to_returnn/converter/converter.py
```diff
@@ -156,6 +157,26 @@ def _make_tf_feed_dict(self, input: Data):
...
+      non_deterministic_tensor_entries =  []
+      entries_to_visit = [out_entry]
+      visited_entries = set()
+      while entries_to_visit:
+        entry_ = entries_to_visit.pop(0)
+        if entry_ in visited_entries:
+          continue
+        visited_entries.add(entry_)
+        if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
+          non_deterministic_tensor_entries.append(entry_)
```
albertz, line 170: After this, you should `continue`, and not add further dependencies, right?

pytorch_to_returnn/converter/converter.py
```diff
@@ -168,10 +168,10 @@ def _make_tf_feed_dict(self, input: Data, out_entry: Optional[TensorEntry] = Non
         visited_entries.add(entry_)
         if not all(call_.module.module.is_deterministic for call_ in entry_.output_from_calls):
           non_deterministic_tensor_entries.append(entry_)
+          continue
         for call_ in entry_.output_from_calls:
-          for prev_entry_ in [pe for pe in call_.inputs_args if isinstance(pe, TensorEntry)]:
-            if prev_entry_ not in visited_entries:
-              entries_to_visit.append(prev_entry_)
+          for prev_entry_ in [pe for pe in call_.inputs_tensor_deps if pe not in visited_entries]:
```
albertz, line 173: As said, I would not create such a temp list here. You can simply move any extra checks into the loop.

pytorch_to_returnn/import_wrapper/torch_wrappers/function.py
```diff
@@ -0,0 +1,17 @@
+from .module import WrappedModuleBase
+
+
+class Functional(WrappedModuleBase):
```
vieting, line 4: I wonder if we should rename this to describe it a bit better, e.g. to `WrappedTorchFunction` or so

albertz: Yes, maybe.

albertz (approved):
Looks all good now to me. Let's wait for the tests once more.

PR #92, opened 2022-01-26
add torch.cosine_similarity

vieting:


vieting:
So far, I added a `DotProductAlongDim` module to do the dot product. The rest is done inside `torch.nn.functional.cosine_similarity` using existing functions. However, there is a problem with the `DotLayer`:

```
    line: assert not set(b_reduce_axes).intersection(b_var_axes)
    locals:
      set = <builtin> <class 'set'>
      b_reduce_axes = <local> [2]
      intersection = <not found>
      b_var_axes = <local> [2]
```
for the layer
```
*** root/'DotProductAlongDim' layer dict: {'class': 'dot', 'red1': 'F', 'red2': 'F', 'from': ['data', 'add']}
```
with inputs
```
layer root/'data' output: Data{'data', [B,T|'time:data'[B],F|F'feature:data'(7)]}
layer root/'add' output: Data{'add_output', [B,T|'time:data'[B],F|F'feature:data'(7)]}
```

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -785,6 +785,13 @@ def multi_head_attention_forward(
     return attn_output, None
 
 
+def cosine_similarity(x1: Tensor, x2: Tensor, dim: Optional[int] = 1, eps: Optional[float] = 1e-8) -> Tensor:
+  mod = modules.DotProductAlongDim(dim=dim)
+  numerator = mod(x1, x2)
+  denominator = (x1 ** 2).sum() * (x2 ** 2).sum()
```
albertz, line 791: This is missing `sqrt` on the denominator, right?

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -123,6 +122,41 @@ def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]
     return {"class": "dot", "red1": red1, "red2": red2, "var1": var1, "var2": var2, "from": sources}
 
 
+class DotProductAlongDim(Module):
+  """
+  Maps to RETURNN DotLayer for a special case
+  """
+  is_original_torch_module = False
+
+  def __init__(self, dim: int):
```
albertz, line 131: I would rather call it `reduce_dim` or so.
But also, I wonder whether it is more useful to not have this argument in the `__init__` but directly in the `create_returnn_layer_dict`?

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -123,6 +122,41 @@ def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]
     return {"class": "dot", "red1": red1, "red2": red2, "var1": var1, "var2": var2, "from": sources}
 
 
+class DotProductAlongDim(Module):
```
albertz, line 125: I think I would simply call it `Dot` or `GenericDot`, and allow for potential generalization later when needed.

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -123,6 +122,41 @@ def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]
...
+    assert all(len(input_.shape) >= 2 for input_ in [input1, input2]), "not implemented otherwise"
+
+    shape = input1.shape
+    assert input2.shape == shape, "not implemented otherwise"  # could e.g. consider broadcasting axes
+    del shape[self.dim]
+    shape = tuple(shape)
+    returnn_axis_from_torch_axis = {i: i for i in range(len(shape))}
+    return shape, returnn_axis_from_torch_axis
+
+  def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]:
```
albertz, line 149:
```suggestion
  def create_returnn_layer_dict(self, a: Tensor, b: Tensor, *, reduce_dim_a: int, reduce_dim_b: int) -> Dict[str, Any]:
```

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -123,6 +122,41 @@ def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]
...
+  """
+  Maps to RETURNN DotLayer for a special case
+  """
+  is_original_torch_module = False
+
+  def __init__(self, dim: int):
+    super(DotProductAlongDim, self).__init__()
+    self.dim = dim
+
+  def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
```
albertz, line 135: Do you need the custom `_get_output_shape_from_returnn` for this layer? I would have assumed that the base logic should work.

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -123,6 +122,41 @@ def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]
...
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    assert len(inputs_flat) == 2
+    input1, input2 = inputs_flat
+    assert all(isinstance(input_, Tensor) for input_ in [input1, input2])
+    assert all(len(input_.shape) >= 2 for input_ in [input1, input2]), "not implemented otherwise"
+
+    shape = input1.shape
+    assert input2.shape == shape, "not implemented otherwise"  # could e.g. consider broadcasting axes
+    del shape[self.dim]
```
albertz, line 144: Did you test this code? This here would not work I think because `shape` is a tuple (thus not mutable). You need to convert it to `list` first.

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -122,38 +122,17 @@ def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]
...
-
-  def create_returnn_layer_dict(self, *inputs: Tensor, **kwargs) -> Dict[str, Any]:
-    sources = [self._get_input_layer_name(source) for source in inputs]
-    assert len(sources) == 2
-
-    assert inputs[0].shape == inputs[1].shape
-    dim0 = self._get_input_axis_to_returnn(inputs[0], self.dim)
-    dim1 = self._get_input_axis_to_returnn(inputs[1], self.dim)
-    assert dim0 == dim1
+  def create_returnn_layer_dict(self, a: Tensor, b: Tensor, reduce_dim_a: int, reduce_dim_b: int) -> Dict[str, Any]:
```
albertz, line 131:
```suggestion
  def create_returnn_layer_dict(self, a: Tensor, b: Tensor, *, reduce_dim_a: int, reduce_dim_b: int) -> Dict[str, Any]:
```

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -786,9 +786,8 @@ def multi_head_attention_forward(
 
 
 def cosine_similarity(x1: Tensor, x2: Tensor, dim: Optional[int] = 1, eps: Optional[float] = 1e-8) -> Tensor:
-  mod = modules.DotProductAlongDim(dim=dim)
-  numerator = mod(x1, x2)
-  denominator = (x1 ** 2).sum() * (x2 ** 2).sum()
+  numerator = modules.Dot()(x1, x2, dim, dim)
```
albertz, line 789:
```suggestion
  numerator = modules.Dot()(x1, x2, reduce_dim_a=dim, reduce_dim_b=dim)
```

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -785,6 +785,12 @@ def multi_head_attention_forward(
     return attn_output, None
 
 
+def cosine_similarity(x1: Tensor, x2: Tensor, dim: Optional[int] = 1, eps: Optional[float] = 1e-8) -> Tensor:
+  numerator = modules.Dot()(x1, x2, dim, dim)
+  denominator = sqrt(pow(x1, 2).sum() * pow(x2, 2).sum())
```
albertz, line 790: Maybe using `rqdrt` instead and then multiplying this would be more stable?

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -785,6 +785,12 @@ def multi_head_attention_forward(
     return attn_output, None
 
 
+def cosine_similarity(x1: Tensor, x2: Tensor, dim: Optional[int] = 1, eps: Optional[float] = 1e-8) -> Tensor:
```
albertz, line 788: You don't use the `eps`. This is probably an error. It probably is added to the input of `rsqrt`?

Also, is `eps=None` really allowed? Otherwise `Optional` is wrong.

albertz:
Btw, you need to rebase this such that we get the new behavior version (#93).

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -799,6 +803,12 @@ def multi_head_attention_forward(
...
+  numerator = modules.Dot()(x1, x2, reduce_dim_a=dim, reduce_dim_b=dim)
+  denominator = rsqrt(pow(x1, 2).sum() * pow(x2, 2).sum())
```
albertz, line 808: Maybe call it `inv_denominator` or so.

pytorch_to_returnn/torch/nn/functional.py
vieting: It's not added, but the maximum of the norms and `eps` is taken, see https://pytorch.org/docs/stable/generated/torch.nn.CosineSimilarity.html

vieting: Do you think using `min(cat(...))` makes sense here? Then it has to be fixed for scalars I think.

albertz: Ok, anyway, we should do the same then.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -799,6 +807,12 @@ def multi_head_attention_forward(
     return attn_output, None
 
 
+def cosine_similarity(x1: Tensor, x2: Tensor, dim: Optional[int] = 1, eps: float=1e-8) -> Tensor:
+  numerator = modules.Dot()(x1, x2, reduce_dim_a=dim, reduce_dim_b=dim)
+  inv_denominator = min(cat([rsqrt(pow(x1, 2).sum() * pow(x2, 2).sum()), tensor(eps ** -1)]))
```
albertz, line 812: Use `minimum` instead.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -799,6 +807,12 @@ def multi_head_attention_forward(
     return attn_output, None
 
 
+def cosine_similarity(x1: Tensor, x2: Tensor, dim: Optional[int] = 1, eps: float=1e-8) -> Tensor:
```
albertz, line 810: Is `dim=None` really allowed?

pytorch_to_returnn/torch/nn/functional.py
vieting: To which RETURNN layer would we map that?

albertz: Maybe simply via `EvalLayer` and `sources: [a, b], eval: "tf.minimum(source(0), source(1))"`

vieting:
Apparently the output still does not match.

albertz:
> Apparently the output still does not match.

This is strange. Can you check at what point it gets wrong? You could also calculate it by hand for some simple values.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -799,6 +810,12 @@ def multi_head_attention_forward(
     return attn_output, None
 
 
+def cosine_similarity(x1: Tensor, x2: Tensor, dim: int=1, eps: float=1e-8) -> Tensor:
+  numerator = modules.Dot()(x1, x2, reduce_dim_a=dim, reduce_dim_b=dim)
+  inv_denominator = minimum(rsqrt(pow(x1, 2).sum() * pow(x2, 2).sum()), tensor(eps ** -1))
```
albertz, line 815: The `sum` is wrong. You only want to reduce the `dim`.

albertz (approved):

PR #93, opened 2022-01-26
add RETURNN behavior_version

vieting:
For #92, we need RETURNN's `behavior_version` to have the correct behavior of the `DotLayer`. There might be fixes necessary due to this change.

### DIFF
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -317,6 +317,7 @@ class Converter:
       config = Config({
         "extern_data": {"data": self._returnn_in_data_dict},
         "debug_print_layer_output_template": True,
+        "behavior_version": 12,
       })
       network = TFNetwork(config=config, name="root", train_flag=self.train)
       network.construct_from_dict(self._returnn_net_dict)
@@ -361,6 +362,7 @@ class Converter:
       config = Config({
         "extern_data": {"data": self._returnn_in_data_dict},
         "debug_print_layer_output_template": True,
+        "behavior_version": 12,
       })
       network = TFNetwork(config=config, name="root", train_flag=self.train)
       network.construct_from_dict(net_dict)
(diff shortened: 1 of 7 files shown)


pytorch_to_returnn/converter/converter.py
```diff
@@ -294,6 +294,7 @@ def _run_returnn_standalone_net_dict(self):
       config = Config({
         "extern_data": {"data": self._returnn_in_data_dict},
         "debug_print_layer_output_template": True,
+        "behavior_version": 12,
```
albertz, line 297: I would not hardcode this number, but rather put it somewhere as a global, e.g. to `naming/returnn_ctx.py`.

albertz:
One fix:
Change
```
net_dict={name: {"class": "subnetwork", "from": "data", "subnetwork": {"output": {"class": "copy"}}}}
```
to
```
net_dict={name: {"class": "subnetwork", "from": "data", "subnetwork": {"output": {"class": "copy", "from": "data"}}}}
```

albertz:
Ah we have to change `get_returnn_axis_description`. When you remove the `if dim_tag.dyn_size is not None:` and just always use this code, does this already work?

pytorch_to_returnn/torch/nn/modules/batchnorm.py
```diff
@@ -96,7 +96,7 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
     return {
       "class": "batch_norm", "from": self._get_input_layer_name(input),
       "update_sample_only_in_training": True, "delay_sample_update": True,
-      "param_version": 1,
+      "param_version": 1, "masked_time": True,
```
albertz, line 99: This is actually not the same as PyTorch. In PyTorch, the time is not not masked.

albertz: You might also want to update the param version. This allows to use the efficient fused batch norm op then.

vieting: Right. For the param version, which do you suggest?

albertz: param_version = 2. This will change the param name (it gets the prefix "v2_") and it will not add the broadcast dims to the variable.

albertz:
You also need to adapt `BatchNorm1d.import_params_torch_to_returnn` for the new param version.

vieting:
> out_shape = [layer.output.dim]

That seems to work. So we don't neet `_expand_dims` at all anymore, I guess.

albertz (approved):
Looks all good.

Issue #94, opened 2022-01-26
changed unsqueeze behavior


vieting:
I have a case with a `(B, T)` tensor which is unsqueezed to `(B, 1, T)` (adding feature dim for raw waveform), however, the RETURNN data of the output is `(B, T, F)`, so `F` and `T` are assigned to the wrong axes. It seems that the behavior changed and the RETURNN axes are not handled correctly now (or this just did not become obvious before).

albertz:
I don't see from this description why this is a problem?


vieting ***@***.***> schrieb am Mi., 26. Jan. 2022, 16:49:

vieting:
Yes sorry, that was not clear. The tensor is unsqueezed to `(B, 1, T)`, however, the RETURNN data of the output is `(B, T, F)`, so `F` and `T` are assigned to the wrong axes. I'll update the description.

albertz:
But why is this axis assignment a problem?


vieting ***@***.***> schrieb am Mi., 26. Jan. 2022, 18:14:

vieting:
I have a `ConvLayer` afterwards and now we explicitly specify `in_spatial_dims` since #93. In the case here, I get `'in_spatial_dims': ['F']` and `AssertionError: invalid in_spatial_dims [Dim{'time:data'[B]}]`

vieting:
The `SplitDimsLayer` is called with `{'class': 'split_dims', 'from': 'data', 'axis': 'T', 'dims': [1, -1]}`. I think it's wrong/unintended, that in `SplitDimsLayer.get_out_data_from_opts` shortly before the end, `out.time_dim_axis` is set to 1 instead of 2. This corresponds to the splitted dim `1` instead of `-1` which would be the original time dim axis.

albertz:
> The `SplitDimsLayer` is called with `{'class': 'split_dims', 'from': 'data', 'axis': 'T', 'dims': [1, -1]}`. I think it's wrong/unintended, that in `SplitDimsLayer.get_out_data_from_opts` shortly before the end, `out.time_dim_axis` is set to 1 instead of 2. This corresponds to the splitted dim `1` instead of `-1` which would be the original time dim axis.

What is `data` in this case?

But even if this behavior is maybe unexpected (and maybe should be fixed), I still don't understand why it is a problem.

albertz:
> because feature dim axes are not allowd in `in_spatial_dims`.

Why do you think so? No, this should not be the case.

[...]

PR #95, opened 2022-01-26
Add unsqueeze test case

vieting:
The issue described in #94 is visible in the test case added here.

### DIFF
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -1049,6 +1049,27 @@ def test_unsqueeze2():
   verify_torch_and_convert_to_returnn(model_func, inputs=x)
 
 
+def test_unsqueeze3():
+  from pytorch_to_returnn.naming import Naming
+  n_batch, n_time = 3, 7
+  n_in, n_out = 1, 4
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    if typing.TYPE_CHECKING or not wrapped_import:
+      import torch
+    else:
+      torch = wrapped_import("torch")
+    out = inputs.unsqueeze(1)
+    model = torch.nn.Conv1d(in_channels=n_in, out_channels=n_out, kernel_size=2)
+    out = model(out)
+    return out
+
+  rnd = numpy.random.RandomState(42)
+  x = rnd.normal(0., 1., (n_batch, n_time)).astype("float32")
+  verify_torch_and_convert_to_returnn(model_func, inputs=x, inputs_data_kwargs={
+      "shape": (None,), "batch_dim_axis": 0, "time_dim_axis": 1, "feature_dim_axis": None})
+
+
 def test_transpose_unsqueeze():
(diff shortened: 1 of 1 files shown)


tests/test_layers.py
```diff
@@ -1049,6 +1049,25 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
+  from pytorch_to_returnn.naming import Naming
+  n_batch, n_time = 3, 7
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    out = inputs.unsqueeze(1)
+    naming = Naming.get_instance()
+    if naming.import_params_from_torch_namespace:
+      returnn_data = naming.tensors[out].returnn_data
+      assert returnn_data.time_dim_axis == 2
+      assert returnn_data.feature_dim_axis == 1
```
albertz, line 1062: This should not matter here, so you should not test this. This can be arbitrary.

Make a test case for what sth which actually fails.

vieting:
I updated the test case to match exactly the situation that I encountered. https://github.com/rwth-i6/returnn/pull/914 fixed this now, but if you test this with a RETURNN version before, you will get the error that I described. We might still want to add this test case to avoid such bugs in the future.

albertz:
I assume the issue is that we do not pass `in_dim` to `ConvLayer`. `in_dim` specifies the feature dim axis to be used by the `ConvLayer`.

vieting:
> I assume the issue is that we do not pass `in_dim` to `ConvLayer`. `in_dim` specifies the feature dim axis to be used by the `ConvLayer`.

Do you think we should put effort into doing that? I tried it, but it's at least not straight forward because `in_dim` has to be a dim tag and in this case the corresponding `SizeValue` in `input.shape` is created without a dim tag, so `dim_tag` is set to a default `Dim(dimension=x, description="static_dim")` which cannot be matched in the `ConvLayer`.

albertz:
> > I assume the issue is that we do not pass `in_dim` to `ConvLayer`. `in_dim` specifies the feature dim axis to be used by the `ConvLayer`.
> 
> Do you think we should put effort into doing that?

If this turns out to be more complicated and not really needed currently we can also delay it. So let's skip that then for now.

However, we might stumble upon this in other cases at some later time.

> I tried it, but it's at least not straight forward because `in_dim` has to be a dim tag and in this case the corresponding `SizeValue` in `input.shape` is created without a dim tag, so `dim_tag` is set to a default `Dim(dimension=x, description="static_dim")` which cannot be matched in the `ConvLayer`.

Yea, the question is, why is it without dim tag? There is always a dim tag on RETURNN side, including the input. So we should never need that "static_dim" dim tag. But as said, we can later investigate into that.

albertz:
So, this PR can maybe be marked as ready, and we can just add the test case?

vieting:
Sounds good to me.

albertz (approved):

Issue #96, opened 2022-01-27
Merge batch with modified time dim


vieting:
In a case where the time dim is modified (e.g. due to downsampling/padding in a convolution) and we want to merge it with the batch dim, we currently face the problem that the creation of a `PaddedDim` in the `MergeDimsLayer` does not work because it is not possible to get the dim value for the modified time dim.

A demo test case which reproduces the issue looks like this and can also be found in #97:
```
def test_merge_batch_with_modified_time():
  n_in, n_out = 5, 7
  n_batch, n_time = 3, 11

  def model_func(wrapped_import, inputs: torch.Tensor):
    if typing.TYPE_CHECKING or not wrapped_import:
      import torch
    else:
      torch = wrapped_import("torch")

    conv = torch.nn.Conv1d(in_channels=n_in, out_channels=n_out, kernel_size=3, stride=2)
    y = inputs  # (B,F,T)
    y = conv(y)  # (B,F',T')
    y = y.transpose(1, 2).contiguous()  # (B,T',F')
    _, _, fsz = y.shape
    y = y.view(-1, fsz)  # (B*T',F')
    return y

  rnd = numpy.random.RandomState(42)
  x = rnd.normal(0., 1., (n_batch, n_in, n_time)).astype("float32")
  verify_torch_and_convert_to_returnn(model_func, inputs=x)
```

The stack trace looks like this ([via](https://github.com/rwth-i6/pytorch-to-returnn/runs/4965676578)):
```
ERROR: test_layers.test_merge_batch_with_modified_time
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/.local/lib/python3.8/site-packages/nose/case.py", line 198, in TestBase.runTest
    line: self.test(*self.arg)
    locals:
      self = <local> test_layers.test_merge_batch_with_modified_time
      self.test = <local> <function test_merge_batch_with_modified_time at 0x7f3ba0138a60>
      self.arg = <local> ()
  File "/home/runner/work/pytorch-to-returnn/pytorch-to-returnn/tests/test_layers.py", line 1215, in test_merge_batch_with_modified_time
    line: verify_torch_and_convert_to_returnn(model_func, inputs=x)
    locals:
      verify_torch_and_convert_to_returnn = <global> <function verify_torch_and_convert_to_returnn at 0x7f3ba013f5e0>
      model_func = <local> <function test_merge_batch_with_modified_time.<locals>.model_func at 0x7f3b40211d30>
[...]

PR #97, opened 2022-01-27
Test merge batch with modified time dim

vieting:
I add a test case here which demonstrates the issue described in #96.


vieting:
Should be fixed by https://github.com/rwth-i6/returnn/pull/916, I'm re-running the tests to see.

vieting:
All tests pass now. Should we merge the test case anyway?

albertz:
> All tests pass now. Should we merge the test case anyway?

Yea, I think it might be useful anyway.

PR #98, opened 2022-01-27
randint: allow dynamic inputs

vieting:
So far, we only supported `torch.randint` for integer inputs. In this PR, this should be extended to dynamic inputs.

### DIFF
--- a/pytorch_to_returnn/naming/tensor.py
+++ b/pytorch_to_returnn/naming/tensor.py
@@ -12,7 +12,6 @@ from . import call as _call
 
 
 class TensorEntry:
-  # from ..torch._C import SizeValue
   tensor: ref[_types.Tensor]
   returnn_data: Optional[Data] = None
   returnn_axis_from_torch_axis: Optional[Dict[int, int]] = None
--- a/pytorch_to_returnn/naming/tensor.py
+++ b/pytorch_to_returnn/naming/tensor.py
@@ -12,6 +12,7 @@ from . import call as _call
 
 
 class TensorEntry:
+  # from ..torch._C import SizeValue
   tensor: ref[_types.Tensor]
   returnn_data: Optional[Data] = None
   returnn_axis_from_torch_axis: Optional[Dict[int, int]] = None
@@ -21,7 +22,7 @@ class TensorEntry:
   is_param: bool = False
   is_const: bool = False  # e.g. via from_numpy, empty, zeros, etc
   is_input: bool = False  # in TF1 terminology, would be a placeholder
-  is_dim: Optional[DimensionTag] = None
+  is_size_value: Optional[SizeValue] = None
   output_from_modules: List[_module.ModuleEntry]
   output_from_calls: List[_call.CallEntry]
   parent_owning_modules: List[Tuple[_module.ModuleEntry, str]]  # e.g. param or buffer
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -35,7 +35,9 @@ class Range(Module):
     limit_dim = None
     if isinstance(limit, Tensor):
       assert limit.is_defined
-      limit_dim = limit.returnn_naming_entry.is_size_value.dim_tag
+      limit_dim = None
+      if limit.returnn_naming_entry.is_size_value is not None:
+        limit_dim = limit.returnn_naming_entry.is_size_value.dim_tag
       limit = limit.numpy()
     from .shape import SizeValue
     size = SizeValue((int(limit) - int(start)) // int(delta))
@@ -61,14 +63,15 @@ class RandInt(Module):
     naming = Naming.get_instance()
     call = naming.module_call_stack[-1]
     assert call.module.module is self
-    source = set()
+    source = []
     for sz in size:
       if isinstance(sz, Tensor):
-        assert naming.tensors[sz].is_size_value is not None
-        originating_tensor = naming.tensors[sz].is_size_value.originating_tensor
+        tensor_entry = naming.tensors[sz]
+        assert tensor_entry.is_size_value is not None
+        originating_tensor = tensor_entry.is_size_value.originating_tensor
         if originating_tensor is not None:
-          source.add(self._get_input_layer_name(naming.tensors[sz].is_size_value.originating_tensor))
           if originating_tensor not in call.inputs_tensor_deps:
+            source.append(self._get_input_layer_name(tensor_entry.is_size_value.originating_tensor))
             # add dependency to get complete in feed_dict in Module.make_output_tensor_from_returnn
             call.inputs_tensor_deps.append(naming.tensors[originating_tensor])
     size = tuple(naming.tensors[sz].is_size_value.dim_tag if isinstance(sz, Tensor) else sz for sz in size)
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -61,13 +61,16 @@ class RandInt(Module):
     naming = Naming.get_instance()
     call = naming.module_call_stack[-1]
     assert call.module.module is self
-    # add dependency to get complete in feed_dict in Module.make_output_tensor_from_returnn
-    call.inputs_tensor_deps.append(naming.inputs[0])
     source = set()
     for sz in size:
       if isinstance(sz, Tensor):
-        if naming.tensors[sz].is_size_value.originating_tensor is not None:
+        assert naming.tensors[sz].is_size_value is not None
+        originating_tensor = naming.tensors[sz].is_size_value.originating_tensor
+        if originating_tensor is not None:
           source.add(self._get_input_layer_name(naming.tensors[sz].is_size_value.originating_tensor))
+          if originating_tensor not in call.inputs_tensor_deps:
+            # add dependency to get complete in feed_dict in Module.make_output_tensor_from_returnn
+            call.inputs_tensor_deps.append(naming.tensors[originating_tensor])
     size = tuple(naming.tensors[sz].is_size_value.dim_tag if isinstance(sz, Tensor) else sz for sz in size)
     assert not None in size
     d = {"class": "rand_int", "shape": size, "maxval": high, "minval": low, "dtype": dtype}
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -35,7 +35,7 @@ class Range(Module):
     limit_dim = None
     if isinstance(limit, Tensor):
       assert limit.is_defined
-      limit_dim = limit.returnn_naming_entry.is_dim
+      limit_dim = limit.returnn_naming_entry.is_size_value.dim_tag
       limit = limit.numpy()
     from .shape import SizeValue
     size = SizeValue((int(limit) - int(start)) // int(delta))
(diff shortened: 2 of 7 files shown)


vieting:
Right now, this crashes because in `CallEntry.apply_call`, the module cannot be found in this line `torch_mod = naming.import_params_from_torch_namespace.get_module_by_abs_id_name(mod_abs_name)`. It should be found by the logic that we added to `get_module_by_abs_id_name` in #90 for `WrappedTorchFunction` like `RandInt` here. Here, however, `part_name="mul_randint"` instead of `"randint"`, so it is not contained in `mod_.module.func_name="pytorch_to_returnn.import_wrapper._torch_traced.torch.randint"` and therefore no module is matched. Do you know how to generalize here?

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -52,30 +52,32 @@ class RandInt(Module):
 
   def create_returnn_layer_dict(self, low, high, size, dtype=None) -> Dict[str, Any]:
     dtype = dtype or "int64"
-    assert isinstance(low, int)
-    assert isinstance(high, int)
-    assert all(isinstance(sz, int) for sz in size)
+    low = self._to_int(low)
+    high = self._to_int(high)
+    size = tuple(self._to_int(sz) for sz in size)
```
albertz, line 57: This is wrong. You do not want to cast a tensor to an `int`. You want to pass a tensor (i.e. using `self._get_input_layer_name`), or if it is an `int`, then you can pass it as it is.

vieting: Yes, right. I think also `_get_output_shape_from_returnn` will need to be fixed, but still there is the issue in `get_module_by_abs_id_name`.

vieting:
I added updates based on the discussion in https://github.com/rwth-i6/returnn/pull/919. We cannot directly access `SizeValue.originating_tensor`, so the logic is a bit more involved. Also, we need to add the dependency for the converter as well (`call.inputs_tensor_deps`), as the `feed_dict` for `Module.make_output_tensor_from_returnn` will not be complete otherwise.

vieting:
For me locally, the test passes now. I'm not yet sure why it fails here (current RETURNN master, python3.8, torch 1.8.1, tf 2.3.1...)

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -52,30 +52,51 @@ class RandInt(Module):
...
+    if isinstance(high, Tensor):
+      high = high.type(dtype)
+      high = self._get_input_layer_name(high)
+    naming = Naming.get_instance()
+    call = naming.module_call_stack[-1]
+    assert call.module.module is self
+    source = set()
+    for sz in size:
+      if isinstance(sz, Tensor):
+        for dep_tensor in naming.tensors[sz].creation_stack_call.inputs_flat:
```
albertz, line 67: You should not access `creation_stack_call` here. You should not use any such assumptions.

You should not use `inputs_flat`.

Why do you do that anyway?

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -52,30 +52,51 @@ class RandInt(Module):
 
   def create_returnn_layer_dict(self, low, high, size, dtype=None) -> Dict[str, Any]:
     dtype = dtype or "int64"
-    assert isinstance(low, int)
-    assert isinstance(high, int)
-    assert all(isinstance(sz, int) for sz in size)
-    return {"class": "rand_int", "shape": size, "maxval": high, "minval": low, "dtype": dtype}
+    if isinstance(low, Tensor):
+      low = low.type(dtype)
+      low = self._get_input_layer_name(low)
```
albertz, line 57: Remove this, and use `tensor_entry.is_dim` instead.

albertz:
> I added updates based on the discussion in [rwth-i6/returnn#919](https://github.com/rwth-i6/returnn/pull/919). We cannot directly access `SizeValue.originating_tensor`, so the logic is a bit more involved.

Why?

vieting:
> > I added updates based on the discussion in [rwth-i6/returnn#919](https://github.com/rwth-i6/returnn/pull/919). We cannot directly access `SizeValue.originating_tensor`, so the logic is a bit more involved.
> 
> Why?

Because if a `SizeValue` is input to this module, we get `x.get_tensor()` here (see [here](https://github.com/rwth-i6/pytorch-to-returnn/blob/878f62db88bbb0ed2c362d38f834082f4454dae7/pytorch_to_returnn/naming/naming.py#L205)). From there, it's not straight-forward to retrieve the `SizeValue`, at least not that I'm aware of.

albertz:
> > > I added updates based on the discussion in [rwth-i6/returnn#919](https://github.com/rwth-i6/returnn/pull/919). We cannot directly access `SizeValue.originating_tensor`, so the logic is a bit more involved.
> > 
> > Why?
> 
> Because if a `SizeValue` is input to this module, we get `x.get_tensor()` here (see [here](https://github.com/rwth-i6/pytorch-to-returnn/blob/878f62db88bbb0ed2c362d38f834082f4454dae7/pytorch_to_returnn/naming/naming.py#L205)). From there, it's not straight-forward to retrieve the `SizeValue`, at least not that I'm aware of.

But that should be straight-forward. We could simply change the `is_dim` to the `SizeValue` directly. Or maybe add a `is_size_value`. Or so.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: The `RandIntLayer` does not accept `Dim` for `maxval` or `minval`. Not sure if this would be helpful.

albertz: Ah sorry, I confused that here.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -52,30 +52,50 @@ class RandInt(Module):
...
+    if isinstance(high, Tensor):
+      high = high.type(dtype)
+      high = self._get_input_layer_name(high)
+    naming = Naming.get_instance()
+    call = naming.module_call_stack[-1]
+    assert call.module.module is self
+    # add dependency to get complete in feed_dict in Module.make_output_tensor_from_returnn
+    call.inputs_tensor_deps.append(naming.inputs[0])
```
albertz, line 65: You should not add `naming.inputs[0]` here but the `originating_tensor` below.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -52,30 +52,50 @@ class RandInt(Module):
...
+    source = set()
+    for sz in size:
+      if isinstance(sz, Tensor):
+        if naming.tensors[sz].is_size_value.originating_tensor is not None:
```
albertz, line 69: There should be some error checking such as checking that `naming.tensors[sz].is_size_value` is set.

pytorch_to_returnn/torch/nn/modules/shape.py
```diff
@@ -399,8 +399,8 @@ def _convert_dim_returnn(x: Union[SizeValue, int, Tensor]) -> Union[int, Dim]:
     return int(x)
   if isinstance(x, Tensor):
     tensor_entry = naming.tensors[x]
-    assert x.is_defined and tensor_entry.is_const and tensor_entry.is_dim
-    return tensor_entry.is_dim
+    assert x.is_defined and tensor_entry.is_const and tensor_entry.is_size_value.dim_tag
```
albertz, line 402:
```suggestion
    assert x.is_defined and tensor_entry.is_const and tensor_entry.is_size_value and tensor_entry.is_size_value.dim_tag
```

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -61,13 +61,16 @@ def create_returnn_layer_dict(self, low, high, size, dtype=None) -> Dict[str, An
     naming = Naming.get_instance()
     call = naming.module_call_stack[-1]
     assert call.module.module is self
-    # add dependency to get complete in feed_dict in Module.make_output_tensor_from_returnn
-    call.inputs_tensor_deps.append(naming.inputs[0])
     source = set()
     for sz in size:
       if isinstance(sz, Tensor):
-        if naming.tensors[sz].is_size_value.originating_tensor is not None:
+        assert naming.tensors[sz].is_size_value is not None
```
albertz, line 67: I think the code becomes easier if you first assign `naming.tensors[sz]` to another var, like:
```
tensor_entry = naming.tensors[sz]
```

pytorch_to_returnn/naming/tensor.py
```diff
@@ -12,6 +12,7 @@
 
 
 class TensorEntry:
+  # from ..torch._C import SizeValue
```
albertz, line 15: Do not add commented-out code.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -35,7 +35,7 @@ def _get_output_shape_from_returnn(self,
     limit_dim = None
     if isinstance(limit, Tensor):
       assert limit.is_defined
-      limit_dim = limit.returnn_naming_entry.is_dim
+      limit_dim = limit.returnn_naming_entry.is_size_value.dim_tag
```
albertz, line 38: This is not the same as before. This would crash if `limit.returnn_naming_entry.is_size_value is None`.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -52,30 +52,53 @@ class RandInt(Module):
...
+    if isinstance(low, Tensor):
+      low = low.type(dtype)
+      low = self._get_input_layer_name(low)
+    if isinstance(high, Tensor):
+      high = high.type(dtype)
+      high = self._get_input_layer_name(high)
+    naming = Naming.get_instance()
+    call = naming.module_call_stack[-1]
+    assert call.module.module is self
+    source = set()
```
albertz, line 64: Do not use `set` here. This makes it non-deterministic.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -52,30 +52,53 @@ class RandInt(Module):
...
+    for sz in size:
+      if isinstance(sz, Tensor):
+        assert naming.tensors[sz].is_size_value is not None
+        originating_tensor = naming.tensors[sz].is_size_value.originating_tensor
+        if originating_tensor is not None:
+          source.add(self._get_input_layer_name(naming.tensors[sz].is_size_value.originating_tensor))
```
albertz, line 70: Just move this code inside the next `if`.

pytorch_to_returnn/torch/nn/modules/operator.py
albertz: Use `[]` instead.

vieting:
> For me locally, the test passes now. I'm not yet sure why it fails here (current RETURNN master, python3.8, torch 1.8.1, tf 2.3.1...)

I still don't get why this behaves differently. Especially, as it seems not related to external dependencies.

albertz:
> > For me locally, the test passes now. I'm not yet sure why it fails here (current RETURNN master, python3.8, torch 1.8.1, tf 2.3.1...)
> 
> I still don't get why this behaves differently. Especially, as it seems not related to external dependencies.

I don't know. What if you run it like it is run like here, using `nosetests`?

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -52,30 +54,53 @@ class RandInt(Module):
...
+    call = naming.module_call_stack[-1]
+    assert call.module.module is self
+    source = []
+    for sz in size:
+      if isinstance(sz, Tensor):
+        tensor_entry = naming.tensors[sz]
+        assert tensor_entry.is_size_value is not None
+        for originating_tensor in tensor_entry.is_size_value.get_originating_tensors():
+          if naming.tensors[originating_tensor] not in call.inputs_tensor_deps:
+            source.append(self._get_input_layer_name(tensor_entry.is_size_value.originating_tensor))
```
albertz, line 73:
```suggestion
            source.append(self._get_input_layer_name(originating_tensor))
```

vieting:
> > > For me locally, the test passes now. I'm not yet sure why it fails here (current RETURNN master, python3.8, torch 1.8.1, tf 2.3.1...)
> > 
> > 
> > I still don't get why this behaves differently. Especially, as it seems not related to external dependencies.
> 
> I don't know. What if you run it like it is run like here, using `nosetests`?

It's related to the fact that I just ran the single test. When running `test_randint` and `test_randint_dynamic`, I get the same error.

vieting:
> > > > For me locally, the test passes now. I'm not yet sure why it fails here (current RETURNN master, python3.8, torch 1.8.1, tf 2.3.1...)
> > > 
> > > 
> > > I still don't get why this behaves differently. Especially, as it seems not related to external dependencies.
> > 
> > 
> > I don't know. What if you run it like it is run like here, using `nosetests`?
> 
> It's related to the fact that I just ran the single test. When running `test_randint` and `test_randint_dynamic`, I get the same error.

In both tests, `torch.randint` is a `WrappedTorchFunction` during the traced run. However, in the second test (no matter the order), `Naming.get_instance().modules` is empty because `WrappedIndirectModule.__getattr__` with `item="randint"` is never called, which would subsequently create the entries in `naming.modules`.

vieting:
Thanks for the fix! From my perspective, the PR is ready then.

albertz (approved):

PR #99, opened 2022-02-02
fix Range._get_output_shape_from_returnn

vieting:
In `Range._get_output_shape_from_returnn`, if `limit` is a `Tensor` we currently set the `dim_tag` for the `SizeValue` in `torch_shape`, but not the `originating_tensor`. This might be problematic as can be seen in the testcase that I added

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -37,7 +37,7 @@ class Range(Module):
       assert limit.is_defined
       size = SizeValue((int(limit.numpy()) - int(start)) // int(delta))
       if limit.returnn_naming_entry.is_size_value is not None:
-        size.dim_tag = limit.returnn_naming_entry.is_size_value.dim_tag
+        size.dim_tag = (limit.returnn_naming_entry.is_size_value.dim_tag - int(start)) // int(delta)
         size.originating_tensor = limit.returnn_naming_entry.is_size_value.originating_tensor
     else:
       size = SizeValue((int(limit) - int(start)) // int(delta))
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -33,14 +33,15 @@ class Range(Module):
                                      ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
     from .shape import SizeValue
     limit, start, delta, *_ = inputs_flat
+    limit_size = None
     if isinstance(limit, Tensor):
       assert limit.is_defined
-      size = SizeValue((int(limit.numpy()) - int(start)) // int(delta))
-      if limit.returnn_naming_entry.is_size_value is not None:
-        size.dim_tag = (limit.returnn_naming_entry.is_size_value.dim_tag - int(start)) // int(delta)
-        size.originating_tensor = limit.returnn_naming_entry.is_size_value.originating_tensor
-    else:
-      size = SizeValue((int(limit) - int(start)) // int(delta))
+      limit_size = limit.returnn_naming_entry.is_size_value
+      limit = limit.numpy()
+    size = SizeValue((int(limit) - int(start)) // int(delta))
+    if limit_size is not None:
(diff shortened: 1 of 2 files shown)


pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -31,18 +31,16 @@ def create_returnn_layer_dict(self, limit, start, delta, dtype, sparse=False) ->
...
       assert limit.is_defined
-      limit_dim = None
+      size = SizeValue((int(limit.numpy()) - int(start)) // int(delta))
       if limit.returnn_naming_entry.is_size_value is not None:
-        limit_dim = limit.returnn_naming_entry.is_size_value.dim_tag
-      limit = limit.numpy()
-    from .shape import SizeValue
-    size = SizeValue((int(limit) - int(start)) // int(delta))
-    if limit_dim:
-      size.dim_tag = (limit_dim - int(start)) // int(delta)
```
albertz, line 45: Why did you change the old code so much?
Wouldn't it have been enough to just add `size.originating_tensor = ...` here?

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -31,18 +31,16 @@ def create_returnn_layer_dict(self, limit, start, delta, dtype, sparse=False) ->
...
+        size.dim_tag = limit.returnn_naming_entry.is_size_value.dim_tag
```
albertz, line 40: This is wrong. It should be `(limit_dim - int(start)) // int(delta)` like it was before.

pytorch_to_returnn/torch/nn/modules/operator.py
vieting: I found this 
```
limit_dim = None
...
if limit_dim:
...
```
unnecessarily long and harder to read, especially if we do the same for `originating_tensor` now. But if you prefer, I can also add `originating_tensor` in the same way.

albertz: Why so complicated? Instead of `limit_dim`, you could also directly keep the `size_value`.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -31,18 +31,16 @@ def create_returnn_layer_dict(self, limit, start, delta, dtype, sparse=False) ->
...
-        limit_dim = limit.returnn_naming_entry.is_size_value.dim_tag
-      limit = limit.numpy()
-    from .shape import SizeValue
-    size = SizeValue((int(limit) - int(start)) // int(delta))
-    if limit_dim:
-      size.dim_tag = (limit_dim - int(start)) // int(delta)
+        size.dim_tag = (limit.returnn_naming_entry.is_size_value.dim_tag - int(start)) // int(delta)
+        size.originating_tensor = limit.returnn_naming_entry.is_size_value.originating_tensor
+    else:
+      size = SizeValue((int(limit) - int(start)) // int(delta))
```
albertz, line 43: This code is redundant. I think you should not repeat the code `size = SizeValue(...)` twice.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -31,18 +31,16 @@ def create_returnn_layer_dict(self, limit, start, delta, dtype, sparse=False) ->
...
-      limit_dim = None
+      size = SizeValue((int(limit.numpy()) - int(start)) // int(delta))
       if limit.returnn_naming_entry.is_size_value is not None:
-        limit_dim = limit.returnn_naming_entry.is_size_value.dim_tag
-      limit = limit.numpy()
-    from .shape import SizeValue
-    size = SizeValue((int(limit) - int(start)) // int(delta))
-    if limit_dim:
-      size.dim_tag = (limit_dim - int(start)) // int(delta)
+        size.dim_tag = (limit.returnn_naming_entry.is_size_value.dim_tag - int(start)) // int(delta)
```
albertz, line 40: Do not access always `limit.returnn_naming_entry.is_size_value`. This makes it hard to read. First do sth like
```
limit_size = limit.returnn_naming_entry.is_size_value
```

albertz (approved):

Issue #100, opened 2022-02-02
Adding randint and arange


vieting:
There is currently an isue when addint `randint` and `arange` outputs in the way described below. The error occurs in `BinaryOperator.create_returnn_layer_dict` in `_unify_tensor_axes_returnn_meta`, but I suspect the problem is that the output of `arange` is marked as time dim in the RETURNN `Data` object where it should actually be a batch dim.

I'll add a PR with the test case and post the stack trace here.

A demo test case which reproduces the issue looks like this:

```
def test_arange_dyn_unsqueeze_add():
  n_batch, n_feat = 3, 5

  def model_func(wrapped_import, inputs: torch.Tensor):
    if typing.TYPE_CHECKING or not wrapped_import:
      import torch
    else:
      torch = wrapped_import("torch")
    randint = torch.randint(low=0, high=10, size=(inputs.shape[0], 4))
    arange = torch.arange(inputs.shape[0]).unsqueeze(1)
    return randint + arange

  rnd = numpy.random.RandomState(42)
  x = rnd.normal(0., 1., (n_batch, n_feat)).astype("float32")
  verify_torch_and_convert_to_returnn(model_func, inputs=x)
```

albertz:
What is the exact error (including stack trace)?

vieting:
I'll link it once the test in #101 is finished.

vieting:
Stack trace ([via](https://github.com/rwth-i6/pytorch-to-returnn/runs/5037256503?check_suite_focus=true))
```
FAIL: test_layers.test_arange_dyn_unsqueeze_add
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/.local/lib/python3.8/site-packages/nose/case.py", line 198, in TestBase.runTest
    line: self.test(*self.arg)
    locals:
      self = <local> test_layers.test_arange_dyn_unsqueeze_add
      self.test = <local> <function test_arange_dyn_unsqueeze_add at 0x7f2588b15310>
      self.arg = <local> ()
  File "/home/runner/work/pytorch-to-returnn/pytorch-to-returnn/tests/test_layers.py", line 1429, in test_arange_dyn_unsqueeze_add
    line: verify_torch_and_convert_to_returnn(model_func, inputs=x)
    locals:
      verify_torch_and_convert_to_returnn = <global> <function verify_torch_and_convert_to_returnn at 0x7f2588b10700>
[...]

PR #101, opened 2022-02-02
Test adding randint and arange

vieting:
I add a test case here which demonstrates the issue described in #100.


vieting:
The issue was fixed by https://github.com/rwth-i6/returnn/pull/932. We can merge the test anyway if you like.

Issue #105, opened 2022-02-07
Floordiv for SizeValue


vieting:
When `stride > 1` in the convolution in `test_negative_sampling`, the `SizeValue` for the `torch_shape` in `_ConvNd._get_output_shape_from_returnn` is computed using floordiv. This currently fails. 

I'll make this change to the test in a PR and post log and stack trace here.

albertz:
You mean, it should be ceildiv instead, right?

vieting:
No, I think using floordiv is correct. It just currently results in `ValueError: unknown op kind 'floordiv_right'` in `Dim.complete_dyn_size`. I'll post the full error once the tests in #106 finish.

vieting:
See tests in #106 [here](https://github.com/rwth-i6/pytorch-to-returnn/runs/5090649076?check_suite_focus=true)

Log

<details>

```
>>> Running with standard reference imports...

>>> Running with wrapped imports, wrapping original PyTorch...
*** func call pytorch_to_returnn.import_wrapper._torch_traced.torch._jit_internal._copy_to_script_wrapper(...)
*** func call pytorch_to_returnn.import_wrapper._torch_traced.torch._jit_internal._overload_method(...)
*** func call pytorch_to_returnn.import_wrapper._torch_traced.torch.empty(...)
*** func call pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.init.kaiming_uniform_(...)
*** func call pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.init._calculate_fan_in_and_fan_out(...)
*** func call pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.init.uniform_(...)
*** torch module call pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.modules.conv.Conv1d(...)(...)
*** func call pytorch_to_returnn.import_wrapper._torch_traced.torch.nn.functional.conv1d(...)
*** func call pytorch_to_returnn.import_wrapper._torch_traced.torch.arange(...)
*** func call pytorch_to_returnn.import_wrapper._torch_traced.torch.cat(...)
>>>> Module naming hierarchy:
.tmp_root: (hidden, empty)
Conv1d: <ModuleEntry Conv1d(7, 7, kernel_size=(2,), stride=(3,))> -> ...
[...]

PR #106, opened 2022-02-07
Fix floordiv for SizeValue

vieting:
See #105.

### DIFF
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -48,7 +48,7 @@ def test_randint_dynamic():
 
 
 def test_negative_sampling():
-  n_batch, n_time, n_feat = 3, 5, 7  # B, T', F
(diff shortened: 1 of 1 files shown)


tests/test_layers.py
```diff
@@ -48,15 +48,15 @@ def model_func(wrapped_import, inputs: torch.Tensor):
 
 
 def test_negative_sampling():
-  n_batch, n_time, n_feat = 3, 5, 7  # B, T', F
+  n_batch, n_time, n_feat = 3, 15, 7  # B, T', F
```
albertz, line 51: It's probably more interesting to have a test case where `n_time % stride != 0`.

vieting:
The check passes now thanks to the fix on RETURNN side (https://github.com/rwth-i6/returnn/commit/562fec4da299a34618824af94608bf663ea189cc). We might still want to merge the change as it is closer to the case we actually want to test.

PR #107, opened 2022-02-08
Broadcasting for cosine similarity in negative sampling

vieting:
In a contrastive loss, there is typically a cosine similarity applied after negative sampling. I added this to the test case that already exists, but currently this doesn't work yet because our `torch.cosine_similarity` doesn't support broadcasting. I added it in a commit here. `Dot._get_output_shape_from_returnn` is similar to `Matmul._get_output_shape_from_returnn`.

Then, there is some other issue inside `torch.cosine_similarity` which is also related to the broadcasting in this case and needs to be fixed. I'll describe it below.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/linear.py
+++ b/pytorch_to_returnn/torch/nn/modules/linear.py
@@ -129,11 +129,53 @@ class Dot(Module):
   is_original_torch_module = False
 
   def create_returnn_layer_dict(self, a: Tensor, b: Tensor, *, reduce_dim_a: int, reduce_dim_b: int) -> Dict[str, Any]:
-    assert a.shape == b.shape
+    from .operator import _unify_tensor_axes_returnn_meta
+    (a, b), _ = _unify_tensor_axes_returnn_meta(a, b)
     dim0 = self._get_input_axis_to_returnn(a, reduce_dim_a)
     dim1 = self._get_input_axis_to_returnn(a, reduce_dim_b)
     sources = [self._get_input_layer_name(source) for source in [a, b]]
     return {"class": "dot", "red1": dim0, "red2": dim1, "from": sources}
+
+  def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
+                                     ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
+    assert len(inputs_flat) == 4
+    a, b, red_dim_a, red_dim_b = inputs_flat
+    assert all(isinstance(input_, Tensor) for input_ in [a, b])
+    assert all(len(input_.shape) >= 2 for input_ in [a, b]), "not implemented otherwise"
+    assert a.shape[red_dim_a] == b.shape[red_dim_b]
+
+    max_len = max(a.ndim, b.ndim)
+    shape1 = [None] * (max_len - a.ndim) + list(a.shape)
+    shape2 = [None] * (max_len - b.ndim) + list(b.shape)
+    if red_dim_a < 0:
+      red_dim_a += max_len
+    if red_dim_b < 0:
+      red_dim_b += max_len
+    assert red_dim_a == red_dim_b, "not implemented otherwise"
+    shape = []
+    for ax in range(max_len):
+      if ax == red_dim_a:
+        continue
+      elif shape1[ax] == shape2[ax]:
+        shape.append(shape1[ax])
+      elif shape1[ax] is not None:
+        assert shape2[ax] in [1, None], f"dimensions are not broadcastable: {a.shape} vs. {b.shape}"
+        shape.append(shape1[ax])
+      elif shape2[ax] is not None:
+        assert shape1[ax] in [1, None], f"dimensions are not broadcastable: {a.shape} vs. {b.shape}"
+        shape.append(shape2[ax])
+
+    torch_shape = tuple(shape)
+    num_unmatched_axes = max_len - min(a.ndim, b.ndim)
+    out_len = len(torch_shape)
+    returnn_axis_from_torch_axis = {}
+    # DotLayer moves unmatched axes to the end
+    for ax in range(out_len):
+      if ax < num_unmatched_axes:
+        returnn_axis_from_torch_axis[ax] = out_len - num_unmatched_axes + ax
+      else:
+        returnn_axis_from_torch_axis[ax] = ax - num_unmatched_axes
+    return torch_shape, returnn_axis_from_torch_axis
 
 
 __all__ = [
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -67,9 +67,10 @@ def test_negative_sampling():
     y = inputs.view(-1, fsz)  # (B,T,F) => (B*T,F)
     negs = y[neg_idxs.view(-1)]  # (B*T*N,F)
     negs = negs.view(bsz, tsz, n_negatives, fsz).permute(2, 0, 1, 3)  # to (N,B,T,F)
-    inputs = inputs.unsqueeze(0)  # (1,B,T,F)
-    targets = torch.cat([inputs, negs], dim=0)  #(N+1,B,T,F)
-    return targets
+    inputs_unsqueeze = inputs.unsqueeze(0)  # (1,B,T,F)
+    targets = torch.cat([inputs_unsqueeze, negs], dim=0)  # (N+1,B,T,F)
+    logits = torch.cosine_similarity(inputs.float(), targets.float(), dim=-1).type_as(inputs)
+    return logits
 
   rnd = numpy.random.RandomState(42)
   x = rnd.normal(0., 1., (n_batch, n_time, n_feat)).astype("float32")


vieting:
The remaining issue is also in `torch.cosine_similarity`, namely when doing `pow(x1, 2).sum(dim=dim) * pow(x2, 2).sum(dim=dim)`. As input `x1` and `x2 we have
```
>>> x1
<Tensor name:? tensor:(B(3),'((time:data+-2)//3)+1'[B](5),'static_dim'(7)(7)) returnn_data:'Transpose_1_output' [B,F|F'Conv1d:channel'(7),T|'Conv1d:conv:s0'[B]] axes {0:0,2:1,1:2}>
>>> x2
<Tensor name:? tensor:('static_dim+static_dim'(11)(11),B(3),'((time:data+-2)//3)+1'[B](5),'static_dim'(7)(7)) returnn_data:'Cat_output' ['Unflatten_3_split_dims0+(static-dim-2)'(11),B,F|F'Conv1d:channel'(7),T|'Conv1d:conv:s0'[B]] axes {1:1,3:2,2:3,0:0}>
```
`pow().sum()` works as expected, we have
```
>>> pow(x1, 2).sum(dim=dim)
<Tensor name:? tensor:(B(3),'((time:data+-2)//3)+1'[B](5)) returnn_data:'Reduce_output' [B,T|'Conv1d:conv:s0'[B]] axes id>
>>> pow(x2, 2).sum(dim=dim)
<Tensor name:? tensor:('static_dim'(11)(11),B(3),'((time:data+-2)//3)+1'[B](5)) returnn_data:'Reduce_1_output' [F|'Unflatten_3_split_dims0+(static-dim-2)'(11),B,T|'Conv1d:conv:s0'[B]] axes id>
```
but the multiplication of both is
```
>>> pow(x1, 2).sum(dim=dim) * pow(x2, 2).sum(dim=dim)
<Tensor name:? tensor:('((time:data+-2)//3)+1'[B](5),'static_dim'(11)(11),B(3)) returnn_data:'mul_output' [F|'Unflatten_3_split_dims0+(static-dim-2)'(11),B,T|'Conv1d:conv:s0'[B]] axes {0:2,1:0,2:1}>
```
where the output torch shape is wrong (`(5, 11, 3)` instead of `(11, 3, 5)`).

The problem seems to come up because the mapping between torch and RETURNN axes is incorrect in `Module._base_get_output_shape_from_returnn`. In particular, I wonder whether this `break` [here](https://github.com/rwth-i6/pytorch-to-returnn/blob/main/pytorch_to_returnn/torch/nn/modules/module.py#L904) is correct or if there should be another `elif` before. We explicitly have the case where no dim is added/removed, then where a dim is removed, but no handling for the case where a dim is added. Also, if I change the order of the multiplication (so the tensor with the additional dim comes first), the multiplication works and the test passes.

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -129,12 +129,54 @@ class Dot(Module):
   is_original_torch_module = False
 
   def create_returnn_layer_dict(self, a: Tensor, b: Tensor, *, reduce_dim_a: int, reduce_dim_b: int) -> Dict[str, Any]:
-    assert a.shape == b.shape
+    from .operator import _unify_tensor_axes_returnn_meta
+    (a, b), _ = _unify_tensor_axes_returnn_meta(a, b)
```
albertz, line 133: We should define what `Dot` is actually supposed to do.

>   Maps to RETURNN DotLayer for a special case

What does this really mean? What special case? Why only for a special case?

Using `_unify_tensor_axes_returnn_meta` only makes sense when `a` and `b` are supposed to have exactly the same axes, or maybe broadcast dims.

For `Dot`, this should not be needed. `DotLayer` in RETURNN categorizes the dims of `a` and `b` into three types:

- Axes to be reduced (need to be same in `a` and `b`) (broadcasting does not make sense)
- Shared axes, usually the batch axis
- Unique axes

So, e.g. if `a` is [B,T,F] and `b` is [B,F,G], when you reduce axis F, then B is shared and T and G are unique, and you get [B,T,G].

Given the reduced axis, we can automatically infer the shared axes (by just looking which dims occur in both `a` and `b`) and then the remaining axes are unique.

Maybe the shared axes are actually not supposed to be shared. E.g. [B,T,F] * [B,F,T], reduce=F, but you only want to share B, and then get [B,T,T]. To make that explicit, you can also specify `var1=T` and `var2=T`.

In any case, at no point it makes sense that `a` and `b` are exactly the same (`_unify_tensor_axes_returnn_meta`), and also broadcast dims don't make sense.

albertz:
Can you make a new simple test case specifically about the multiplication?

pytorch_to_returnn/torch/nn/modules/linear.py
vieting: > We should define what Dot is actually supposed to do.

Right, we should make that clearer. I added it because I needed it for the dot product in the cosine similarity. You suggested to allow making it more generic at some point. This is not necessary right now, but it probably helps to clarify the use of `Dot`.

vieting:
> Can you make a new simple test case specifically about the multiplication?

I added `test_multiplication_broadcasting`. Here, I assert that the output torch shape is correct which it currently isn't. In the case of the negative sampling, this leads to an error afterwards. We could also add e.g. `out = out + inputs.expand(3, n_batch, n_feature, n_time)` to get exactly the same error but the underlying problem is the same.

pytorch_to_returnn/torch/nn/modules/linear.py
albertz: Yes, but again, as I said, there should never be the need for `_unify_tensor_axes_returnn_meta` or for broadcast dims.

vieting: Right, I think I can just remove it.

albertz: However, it's probably a good idea when we get explicit on `var1` and `var2`.

vieting: You mean setting the non-matching axes explicitly as `var1` and `var2`?

albertz: But what are non-matching axes? As I said, we first should define what this `Dot` layer is actually supposed to do (purely on PyTorch level, independent of RETURNN).

What are our current use cases?

albertz:
> What do you think about the remaining multiplication issue? We could also override `BinaryOperator._get_output_shape_from_returnn` if it's unclear how to fix the general one.

I wonder that the general code fails here. I think we should fix the general code.

Maybe in the general code, we can order the inputs by ndim. That would fix it here.

pytorch_to_returnn/torch/nn/modules/linear.py
vieting: The only use case is in `torch.cosine_similarity` for the dot product along a specified axis. I'm fine with having this module just for this use case for now.

pytorch_to_returnn/torch/nn/modules/linear.py
```diff
@@ -129,12 +129,52 @@ class Dot(Module):
   is_original_torch_module = False
 
   def create_returnn_layer_dict(self, a: Tensor, b: Tensor, *, reduce_dim_a: int, reduce_dim_b: int) -> Dict[str, Any]:
-    assert a.shape == b.shape
     dim0 = self._get_input_axis_to_returnn(a, reduce_dim_a)
     dim1 = self._get_input_axis_to_returnn(a, reduce_dim_b)
     sources = [self._get_input_layer_name(source) for source in [a, b]]
     return {"class": "dot", "red1": dim0, "red2": dim1, "from": sources}
 
+  def _get_output_shape_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase
```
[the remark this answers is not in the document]
vieting: The order of entries in the torch output shape is wrong. But maybe this is actually the same issue as above.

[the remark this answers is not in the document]
vieting: When I remove this, I end up in `assert len(rem_out) == len(rem_in)  # assumption, otherwise no idea what to do`. The `DotLayer` moves unmatched axes to the end and this is not handled in the general case.

pytorch_to_returnn/torch/nn/modules/linear.py
albertz: I thought we had another use case? We introduced `Dot` already before that, or not?

What exactly is the use case in `torch.cosine_similarity`? What are the inputs `a` and `b`?

vieting: In `torch.cosine_similarity`, we use it to compute the dot product of `a` and `b` along a given axis. `a` and `b` are two tensors which must be broadcastable to a common shape, see https://pytorch.org/docs/stable/generated/torch.nn.functional.cosine_similarity.html

albertz: So this means, what we need for `Dot` is:

- Reduce axis is clear.
- Remove any broadcast dims before. I.e. iterate through the a/b shape and see if one of them is broadcast. If one of them is broadcast, add them to the corresponding `var1/2` of the other.
- The remaining axes should all be shared. But this is implicit when specifying `var1/2`.

The behavior is exactly equivalent to:
```
(a * b).sum(dim=reduce_dim)
```

This should be documented.

vieting: Oh, right. We could just use `(a * b).sum(dim=reduce_dim)` in `torch.cosine_similarity` and remove `Dot` completely. I think this would be the cleanest thing as it's not used anywhere else.

pytorch_to_returnn/torch/nn/functional.py
```diff
@@ -816,7 +816,7 @@ def multi_head_attention_forward(
 
 
 def cosine_similarity(x1: Tensor, x2: Tensor, dim: int=1, eps: float=1e-8) -> Tensor:
-  numerator = modules.Dot()(x1, x2, reduce_dim_a=dim, reduce_dim_b=dim)
+  numerator = (x1 * x2).sum(dim=dim)
```
albertz, line 819: Why did you change this? I think it's more efficient to use the `Dot` module. Unless you don't care about this here?

pytorch_to_returnn/torch/nn/modules/linear.py
albertz: Ah I missed this message. But `(a * b).sum(dim=reduce_dim)` is less efficient, and requires more memory. Esp when the dim to be reduced is large. Maybe not so important here?

pytorch_to_returnn/torch/nn/functional.py
vieting: As I commented above, this is the only use of the `Dot` module. So if it's possible to rewrite and do it without, I thought that'd be cleaner.

albertz (approved):

PR #110, opened 2022-02-17
fix serialization for SizeValues

vieting:
The serialization should get the `int` value of a `SizeValue`, otherwise we'll have something in the output network like `'shape': ('static_dim'(768)(768),)` which will obviously lead to an error.

### DIFF
--- a/pytorch_to_returnn/converter/converter.py
+++ b/pytorch_to_returnn/converter/converter.py
@@ -159,6 +159,7 @@ class Converter:
     config = dim_tags_proxy.collect_dim_tags_and_transform_config(config)
 
     code_lines = [
+      "import numpy\n",
       "from returnn.tf.util.data import Dim, batch_dim, single_step_dim, SpatialDim, FeatureDim\n\n",
       "use_tensorflow = True\n",
       f"behavior_version = {returnn_behavior_version}\n\n",
--- a/tests/returnn_helpers.py
+++ b/tests/returnn_helpers.py
@@ -1,0 +1,106 @@
+"""
+Helpers for RETURNN
+"""
+
+from __future__ import annotations
+from typing import Dict, Tuple, Any
+import returnn.tf.engine
+import returnn.datasets
+
+
+def dummy_run_net(config: Dict[str, Any], *, train: bool = False):
+  """
+  Runs a couple of training iterations using some dummy dataset on the net dict.
+  Use this to validate that the net dict is sane.
+  Note that this is somewhat slow. The whole TF session setup and net construction can take 5-30 secs or so.
+  It is not recommended to use this for every single test case.
+
+  The dummy dataset might change at some point...
+
+  Maybe this gets extended...
+  """
+  from returnn.tf.engine import Engine
+  from returnn.datasets import init_dataset
+  from returnn.config import Config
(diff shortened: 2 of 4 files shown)


albertz:
Some test case which tests the serialization would be good, e.g. just serialization + the eval + dummy run.

vieting:
> Some test case which tests the serialization would be good, e.g. just serialization + the eval + dummy run.

I added a test case. Not sure if the `arange` is the ideal test, but the serialized config contains both a dim tag and a `SizeValue`. Would you add another run based on the serialized network dict, similar to `Converter._run_returnn_standalone_net_dict`?

tests/test_layers.py
```diff
@@ -1770,6 +1770,26 @@ def model_func(wrapped_import, inputs: torch.Tensor):
...
+    arange = torch.arange(inputs.shape[1])
+    return inputs + torch.reshape(arange, (1, -1))
+
+  rnd = numpy.random.RandomState(42)
+  x = rnd.normal(0., 1., (n_batch, n_feat)).astype("float32")
+  converter = verify_torch_and_convert_to_returnn(model_func, inputs=x)
+
+  cfg = converter.get_returnn_config_serialized()
+  print(f"Serialized config:\n\n{cfg}")
+  exec(cfg)
```
albertz, line 1790: I would further extend this to also try to load it again via RETURNN. The result of `exec` can be put into a new `cfg` object. See `config_net_dict_via_serialized` and then `dummy_run_net` in the test cases of returnn-common.

vieting: I added that. We have to adapt the training dataset though. Did you want to do training here also? Or just forwarding as in the other tests?

tests/returnn_helpers.py
```diff
@@ -0,0 +1,106 @@
...
+#     if with_axis:
+#       opts["axis"] = time_dim
+#     out = net(data, **opts, name=name_ctx)
+#     if isinstance(out, tuple):
+#       out = out[0]
+#     assert isinstance(out, nn.Tensor)
+#     out.mark_as_default_output()
+#
+#   config_code = name_ctx.get_returnn_config_serialized()
+#   return config_net_dict_via_serialized(config_code)
```
albertz, line 92: You should never add commented-out code.
Instead, add a comment with a link to other code in returnn-common or so if you think this code can become relevant.

pytorch_to_returnn/converter/converter.py
```diff
@@ -159,6 +159,7 @@ def get_returnn_config_serialized(self) -> str:
     config = dim_tags_proxy.collect_dim_tags_and_transform_config(config)
 
     code_lines = [
+      "import numpy\n",
```
albertz, line 162: Why is this needed?

vieting: If we have a `ConstantLayer`, it potentially contains `numpy.array[...]`.

albertz:
Now it fails because you use `DummyDataset` which creates some data which is not compatible to what you expect here in the test, specifically you expect (B,F) but `DummyDataset` generates (T,F) and then batched as (B,T,F).

Maybe for this serialization test, just make some test which works on (B,T,F) to not make it too complicated.

PR #115, opened 2022-02-25
Output dim tag of torch.full

### LINKED ISSUE #114: Output dim tag of torch.full
In #113, it shows that the dim tag of the output shape of `torch.full` is not set correctly. An example to reproduce the issue looks like this:
```
def test_full_out_dim_tag():
  n_batch, n_feat = 3, 5

  def model_func(wrapped_import, inputs: torch.Tensor):
    if typing.TYPE_CHECKING or not wrapped_import:
      import torch
    else:
      torch = wrapped_import("torch")
    full = torch.full((inputs.shape[0],), 42)
    arange = torch.arange(full .shape[0])
    return arange + full 

  x = rnd.normal(0., 1., (n_batch, n_feat)).astype("float32")
```
which fails in `pytorch_to_returnn/torch/nn/modules/operator.py`, line 594, in `_unify_tensor_axes_returnn_meta` with
```
    line: assert all(dim.dimension == d.dimension for d in dims_for_axis if d is not None and d.dimension != 1), (
            f"invalid input {x} axis {i} dim {dim}")
    locals:
      all = <builtin> <built-in function all>
      dim = <local> Dim{'Range:range'(3)}
      dim.dimension = <local> 3
      d = <not found>
      d.dimension = <not found>
      dims_for_axis = <local> [Dim{'Range:range'(3)}, Dim{B}]
AssertionError: invalid input <Tensor name:? tensor:('static_dim'(3)(3),) returnn_data:'Cast_output' [F|'Range:range'(3)] axes id> axis 0 dim Dim{'Range:range'(3)}
```
(I'll add a full stack trace and log later).

We can see that the dim tag in the output shape of `torch.full` is wrong, but in `returnn_data` it's still correct. When creating a range based on that, the wrong dim tag of the output shape is also used on RETURNN side, which will lead to the error when trying to add both tensors.
```
      arange = <local> <Tensor name:? tensor:('static_dim'(3)(3),) returnn_data:'Range_output' [F|'Range:range'(3)] axes id>
      full = <local> <Tensor name:? tensor:('static_dim'(3)(3),) returnn_data:'FullStatic_const' [B] axes id>
```

vieting:
Fix #114

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -516,6 +516,17 @@ class Reduce(Module):
     axes = [self._get_input_axis_to_returnn(input, axis) for axis in axes]
     return {"class": "reduce", "mode": self.mode, "axes": axes, "from": self._get_input_layer_name(input)}
 
+  def make_output_tensor_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase) -> Tensor:
+    tensor = super(Reduce, self).make_output_tensor_from_returnn(inputs_flat, layer)
+    if self.mode == "max":
+      naming = Naming.get_instance()
+      assert len(inputs_flat) == 1
+      input_entry = naming.tensors[inputs_flat[0]]
+      if input_entry.is_size_value:
+        tensor_entry = naming.tensors[tensor]
+        tensor_entry.is_size_value = input_entry.is_size_value
+    return tensor
+
 
 class Length(Module):
   """
--- a/pytorch_to_returnn/torch/nn/modules/operator.py
+++ b/pytorch_to_returnn/torch/nn/modules/operator.py
@@ -36,11 +36,11 @@ class Range(Module):
     limit, start, delta, *_ = inputs_flat
     size = None
     if isinstance(limit, Tensor):
-      assert limit.is_defined
       limit_size = limit.returnn_naming_entry.is_size_value
       if limit_size is not None:
         size = (limit_size - int(start)) // int(delta)
       else:
+        assert limit.is_defined
         limit = limit.numpy()
     if size is None:
       size = SizeValue((int(limit) - int(start)) // int(delta))
--- a/pytorch_to_returnn/naming/tensor.py
+++ b/pytorch_to_returnn/naming/tensor.py
@@ -19,7 +19,7 @@ class TensorEntry:
   validated_to_torch_tf_feed_dict: Optional[Dict[tf.Tensor, numpy.ndarray]] = None
   validated_to_torch_tf_sizes_feed_dict: Optional[Dict[tf.Tensor, numpy.ndarray]] = None
   is_param: bool = False
-  is_const: bool = False  # e.g. via from_numpy, empty, zeros, etc
+  is_const: bool = False  # e.g. via from_numpy, empty, zeros, etc (without dependency on dynamic axes)
   is_input: bool = False  # in TF1 terminology, would be a placeholder
   is_size_value: Optional[SizeValue] = None
   output_from_modules: List[_module.ModuleEntry]
--- a/pytorch_to_returnn/torch/nn/modules/variable.py
+++ b/pytorch_to_returnn/torch/nn/modules/variable.py
@@ -5,6 +5,7 @@ import tensorflow as tf
 from returnn.tf.layers.basic import LayerBase, VariableLayer
 from typing import Tuple, List, Optional
 from ...tensor import Tensor, dtype as _DType
+from ..._C import SizeValue
 from ..parameter import Parameter
 from .module import Module
 from ....naming import Naming
@@ -138,6 +139,8 @@ class FullStatic(Module):
     entry.returnn_axis_from_torch_axis = {i: i for i in range(tensor.ndim)}
     for dim in range(tensor.ndim):
       tensor.shape[dim].dim_tag = entry.returnn_data.dim_tags[dim]
+    if isinstance(self.fill_value, SizeValue):
+      entry.is_size_value = self.fill_value
     return tensor
 
 
--- a/pytorch_to_returnn/torch/nn/modules/variable.py
+++ b/pytorch_to_returnn/torch/nn/modules/variable.py
@@ -133,9 +133,11 @@ class FullStatic(Module):
     from ..._C import from_numpy
     tensor = from_numpy(numpy.full(size, self.fill_value, dtype=self.dtype.name))
     entry = naming.register_tensor(tensor)
-    entry.is_const = True
+    entry.is_const = not any(isinstance(x, Tensor) for x in inputs_flat)
     entry.returnn_data = layer.output
     entry.returnn_axis_from_torch_axis = {i: i for i in range(tensor.ndim)}
+    for dim in range(tensor.ndim):
+      tensor.shape[dim].dim_tag = entry.returnn_data.dim_tags[dim]
     return tensor
 
 
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -1542,6 +1542,23 @@ def test_full():
   verify_torch_and_convert_to_returnn(model_func, inputs=x)
 
 
+def test_full_out_dim_tag():
+  n_batch, n_feat = 3, 5
+
+  def model_func(wrapped_import, inputs: torch.Tensor):
+    if typing.TYPE_CHECKING or not wrapped_import:
+      import torch
+    else:
+      torch = wrapped_import("torch")
+    full = torch.full((inputs.shape[0],), 42)
+    arange = torch.arange(full.shape[0])
+    return arange + full
+
+  rnd = numpy.random.RandomState(42)
+  x = rnd.normal(0., 1., (n_batch, n_feat)).astype("float32")
+  verify_torch_and_convert_to_returnn(model_func, inputs=x)
+
+
 def test_const_with_batch_and_gather():
   n_batch, n_time = 3, 7
 


albertz:
I don't remember the details anymore. `limit.is_defined` makes only sense for static dims, right? And if we have `limit.returnn_naming_entry.is_size_value`, we don't need that (we don't need `limit = limit.numpy()`).

albertz:
I'm not sure if we have `limit.returnn_naming_entry.is_size_value` here though. Effectively, `limit` is also not the size itself but the max of the seq lengths. Which should be the same though. We probably need some extra logic to set this correctly then. E.g. when you do `torch.max(lengths)`.

vieting:
> I don't remember the details anymore. limit.is_defined makes only sense for static dims, right? And if we have limit.returnn_naming_entry.is_size_value, we don't need that (we don't need limit = limit.numpy()).

Yes, I agree.

> I'm not sure if we have limit.returnn_naming_entry.is_size_value here though. Effectively, limit is also not the size itself but the max of the seq lengths. Which should be the same though. We probably need some extra logic to set this correctly then. E.g. when you do torch.max(lengths).

Right, it's `None` here. I'll look into the `torch.max`.

vieting:
> I'm not sure if we have `limit.returnn_naming_entry.is_size_value` here though. Effectively, `limit` is also not the size itself but the max of the seq lengths. Which should be the same though. We probably need some extra logic to set this correctly then. E.g. when you do `torch.max(lengths)`.

So for `Reduce` with `mode="max"`, we would need to check if we can already know what that is?

albertz:
> > I'm not sure if we have `limit.returnn_naming_entry.is_size_value` here though. Effectively, `limit` is also not the size itself but the max of the seq lengths. Which should be the same though. We probably need some extra logic to set this correctly then. E.g. when you do `torch.max(lengths)`.
> 
> So for `Reduce` with `mode="max"`, we would need to check if we can already know what that is?

What fails now with your change is e.g. `test_arange_from_lengths`. In there, we have this code:
```
batch_dim, time_dim = inputs.shape
lengths = torch.full((batch_dim,), time_dim)
arange = torch.arange(torch.max(lengths))
```
In practice, it is a question where `lengths` comes from. I think so far we never really attached the sequence lengths to the original tensor in any way, or did we?

In this synthetic unrealistic example, we could infer that `torch.max(lengths)` refers to `time_dim` which has the correct `is_size_value`. Is the same actually, we maybe can even return the same tensor object.

We should fix this example such that the tests pass. Which probably should not be too difficult. Or otherwise the tests somehow needs to be rewritten.

The more relevant question however is: How will it look in practice, i.e. where do the `lengths` come from, and how are they connected to `inputs`?

Maybe you could just also set `is_size_value` of `lengths` here in this example (and in general, however you would get `lengths`). This is not just a shape dim (scalar) but the actual lengths but we could treat it in the same way. That would then be easy to detect in `torch.max`. Not sure if this is a good solution or whether this causes other problems elsewhere.

vieting:
I have pushed a fix, but it's very specific. In `Reduce` with `mode in ["max", "min", "mean"]`, it checks if the input comes from `FullStatic` and sets `is_size_value` for the tensor entry if the `fill_value` is a `SizeValue`.

vieting:
> The more relevant question however is: How will it look in practice, i.e. where do the lengths come from, and how are they connected to inputs?

I don't remember the details, but I think the motivation for the test was `test_packed_sequence_2` or, more generally, `torch.nn.utils.rnn.pad_packed_sequence`.

albertz:
> I have pushed a fix, but it's very specific. In `Reduce` with `mode in ["max", "min", "mean"]`, it checks if the input comes from `FullStatic` and sets `is_size_value` for the tensor entry if the `fill_value` is a `SizeValue`.

But it only makes sense for `max`, or not?

albertz:
> > The more relevant question however is: How will it look in practice, i.e. where do the lengths come from, and how are they connected to inputs?
> 
> I don't remember the details, but I think the motivation for the test was `test_packed_sequence_2` or, more generally, `torch.nn.utils.rnn.pack_sequence`.

But also there, the same question: Where do the `lengths` come from?

vieting:
> > I have pushed a fix, but it's very specific. In `Reduce` with `mode in ["max", "min", "mean"]`, it checks if the input comes from `FullStatic` and sets `is_size_value` for the tensor entry if the `fill_value` is a `SizeValue`.
> 
> But it only makes sense for `max`, or not?

If it is `FullStatic`, then all values in the tensor are the same, so max = min = mean, ot not?

vieting:
> > > The more relevant question however is: How will it look in practice, i.e. where do the lengths come from, and how are they connected to inputs?
> > 
> > 
> > I don't remember the details, but I think the motivation for the test was `test_packed_sequence_2` or, more generally, `torch.nn.utils.rnn.pack_sequence`.
> 
> But also there, the same question: Where do the `lengths` come from?

In `pack_sequence` we have
```
    batch_dim = sequences.shape[torch_axis_from_returnn_axis[tensor_entry.returnn_data.batch_dim_axis]]
    time_dim = sequences.shape[torch_axis_from_returnn_axis[tensor_entry.returnn_data.time_dim_axis]]
    lengths = full((batch_dim,), time_dim)
    # batch_first is always True because we assume batch-time-major tensor, see above
    return pack_padded_sequence(sequences, lengths, enforce_sorted=enforce_sorted, batch_first=True)
```
so `lengths` is created by `full`. Then in `pack_padded_sequence`, the `lengths` are converted to `batch_sizes` via `_batch_sizes_from_lengths` which uses `torch.arange(torch.max(lengths))`. So this is pretty much the same as what we test in `test_arange_from_lengths`.

albertz:
> > > I have pushed a fix, but it's very specific. In `Reduce` with `mode in ["max", "min", "mean"]`, it checks if the input comes from `FullStatic` and sets `is_size_value` for the tensor entry if the `fill_value` is a `SizeValue`.
> > 
> > But it only makes sense for `max`, or not?
> 
> If it is `FullStatic`, then all values in the tensor are the same, so max = min = mean, ot not?

I don't understand why this is relevant? Normally you would never use `FullStatic` in practice for the seq lengths. The seq lengths are usually never all the same. So in the general case, only `max` is correct.

albertz:
> In `pack_sequence` we have
> 
> ```
>     batch_dim = sequences.shape[torch_axis_from_returnn_axis[tensor_entry.returnn_data.batch_dim_axis]]
>     time_dim = sequences.shape[torch_axis_from_returnn_axis[tensor_entry.returnn_data.time_dim_axis]]
>     lengths = full((batch_dim,), time_dim)
>     # batch_first is always True because we assume batch-time-major tensor, see above
>     return pack_padded_sequence(sequences, lengths, enforce_sorted=enforce_sorted, batch_first=True)
> ```
> 
> so `lengths` is created by `full`. Then in `pack_padded_sequence`, the `lengths` are converted to `batch_sizes` via `_batch_sizes_from_lengths` which uses `torch.arange(torch.max(lengths))`. So this is pretty much the same as what we test in `test_arange_from_lengths`.

But why is this relevant? This code is obviously wrong, and you would use it in practice. You do not have the same seq lengths in practice.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -516,6 +516,23 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
...
+    if self.mode in ["max", "min", "mean"]:
+      # handle special case where input to Reduce is FullStatic. Here, we might already know the output and in case it
+      # is a SizeValue, we can set it for the TensorEntry
+      naming = Naming.get_instance()
+      assert len(inputs_flat) == 1
+      input_entry = naming.tensors[inputs_flat[0]]
+      from .variable import FullStatic
+      if len(input_entry.output_from_calls) == 1:
+        mod = input_entry.output_from_calls[0].module.module
+        if isinstance(mod, FullStatic):
```
albertz, line 530: You should not have any check for `FullStatic` here. `FullStatic` should be totally irrelevant in this function.

pytorch_to_returnn/torch/nn/modules/operator.py
```diff
@@ -516,6 +516,23 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
     axes = [self._get_input_axis_to_returnn(input, axis) for axis in axes]
     return {"class": "reduce", "mode": self.mode, "axes": axes, "from": self._get_input_layer_name(input)}
 
+  def make_output_tensor_from_returnn(self, inputs_flat: List[Tensor], layer: LayerBase) -> Tensor:
+    tensor = super(Reduce, self).make_output_tensor_from_returnn(inputs_flat, layer)
+    if self.mode in ["max", "min", "mean"]:
+      # handle special case where input to Reduce is FullStatic. Here, we might already know the output and in case it
```
albertz, line 522: You should not have a special case for this.

albertz:
> Maybe you could just also set `is_size_value` of `lengths` here in this example (and in general, however you would get `lengths`). This is not just a shape dim (scalar) but the actual lengths but we could treat it in the same way. That would then be easy to detect in `torch.max`. Not sure if this is a good solution or whether this causes other problems elsewhere.

Did you try this?

I.e. this would be a small extra logic in `FullStatic`.

And in `Reduce`, you would just check for `is_size_value` in case of `max` (and only `max`).

vieting:
> > Maybe you could just also set `is_size_value` of `lengths` here in this example (and in general, however you would get `lengths`). This is not just a shape dim (scalar) but the actual lengths but we could treat it in the same way. That would then be easy to detect in `torch.max`. Not sure if this is a good solution or whether this causes other problems elsewhere.
> 
> Did you try this?
> 
> I.e. this would be a small extra logic in `FullStatic`.
> 
> And in `Reduce`, you would just check for `is_size_value` in case of `max` (and only `max`).

I pushed something, is that what you mean?

albertz:
> I pushed something, is that what you mean?

Yes, looks good.

albertz (approved):

PR #117, opened 2022-03-31
Fix for RETURNN updates

vieting:
I think that some updates of RETURNN break things in the converter. I'll rerun the test cases here so we can see.

PR #118, opened 2022-03-31
fix for axes in GenericPadNd

vieting:
In `GenericPadNd`, the axes are currently converted to RETURNN manually before passing them to `self._get_input_axis_to_returnn`, however, `self._get_input_axis_to_returnn` should receive the torch axes. This is fixed here, which also simplifies the code a bit.


vieting:
We first need to fix #117 and then rebase.

albertz (approved):

PR #123, opened 2022-05-23
added test case for unsqueeze with residual connection

dierkes-j:
I added a test case for unsqueeze followed by a residual connection that currently seems to fail. The problem occured when trying to convert the current fairseq wav2vec2 conformer implementation

### DIFF
--- a/tests/test_layers.py
+++ b/tests/test_layers.py
@@ -1181,6 +1181,40 @@ def test_transpose_unsqueeze():
   verify_torch_and_convert_to_returnn(model_func, inputs=x)
 
 
+def test_unsqueeze_residual():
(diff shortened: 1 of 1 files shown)


albertz:
Is this still relevant? I see that #124 is closed. The test here is failing but maybe it needs just a rebase?

dierkes-j:
> Is this still relevant? I see that #124 is closed. The test here is failing but maybe it needs just a rebase?

Yes, still relevant. I rebased.

dierkes-j:
I think the failing tests are not related to my commit, are they? Same problem as in https://github.com/rwth-i6/pytorch-to-returnn/issues/125?

albertz (approved):

PR #126, opened 2022-09-15
small extensions

vieting:
Add some small extensions which are used in a new model that I'd like to convert.


vieting:
Some test cases are failing because of RETURNN updates, not the changes introduced in this PR. See https://github.com/rwth-i6/pytorch-to-returnn/issues/125

albertz (approved):

Issue #125, opened 2022-09-15
Failing Test Cases


vieting:
There are failing test cases for the last commit on the main branch. Since the commit is only a minor change to the readme, it is very likely that recent updates of RETURNN cause the failures.

vieting:
As this came up again, I had a look at the issues. They occur when running from a network dict, all runs up until there (including the net dict creation) work.

In `RandIntLayer.get_out_data_from_opts()`, the dim tags inferred from `shape` are 
```
>>> dim_tags
[Dim{B}, Dim{'3*time:data'[B]}]
>>> dim_tags[1].dyn_size
<tf.Tensor 'mul_randint/mul:0' shape=(?,) dtype=int32>
```
However, the `dyn_size` is removed in `get_for_batch_ctx`.
```
>>> dim_tags[1].get_for_batch_ctx(batch, ctx).dyn_size
None
```
I'm not familiar enough with the details there to have a clear idea why this is the case. This leads to the error subsequently in `RandIntLayer.__init__()` when calling `.get_dim_value()` on that dim tag. @albertz do you have an idea how to fix this?

The other two failing tests are related.

albertz:
I cannot see the tests anymore. Can you post the relevant exceptions here?

What is `batch` and `ctx` in your example?

Probably some `complete_dyn_size` is missing here.

vieting:
You can currently see the tests [here](https://github.com/rwth-i6/pytorch-to-returnn/actions/runs/3489868241/jobs/5840540146) from the last commit in main.

```
>>> ctx                                                                                                                                                                                                                         
None
>>> batch                                                                                                                                                                                                                       
BatchInfo{B}
>>> vars(batch)                                                                                                                                                                                                                 
{'_descendants_by_beam_name': {},
 '_dim': None,
 '_global_beam_dims_by_beam_name': {},
 '_global_descendants_by_virtual_dims': {(GlobalBatchDim{B},): BatchInfo{B}},
 '_global_padded_dims_by_dim_tag': {},
 '_packed_dims_by_dim_tag': {},
 'base': None,
 'descendants': [],
 'virtual_dims': [GlobalBatchDim{B}]}
```
I cannot see a difference regarding these between the working `_run_torch_returnn_drop_in()` and the failing `_run_returnn_standalone_net_dict()`.

albertz:
Error:
```
ERROR: test_layers.test_randint_dynamic
----------------------------------------------------------------------
[...]

PR #128, opened 2022-11-21
allow functional depthwise conv

vieting:
Allow using groups for `_FunctionalConvNd`. I also added a test for depthwise convolution with stride 2. This revealed some problems on RETURNN side. It should work with https://github.com/rwth-i6/returnn/pull/1227.


vieting:
The failing test cases are those already discussed in #125. The new one I added (and all others dealing with convolution) pass, so this PR is ready for review.

PR #129, opened 2022-11-23
allow custom modules with custom RETURNN layer

vieting:
I'd like to create a module which has a torch forward function, but nevertheless is mapped to a custom RETURNN layer call. The basic idea is to have something like
```
class MyModule(torch.nn.Module):

  def forward(self, x):
    y = foo(x)
    return y

  def create_returnn_layer_dict(self, input):
    return {"class": "<some_layer>", "from": self._get_input_layer_name(input)}
```

Right now, the custom `create_returnn_layer_dict` is ignored and the code inside `forward` is wrapped. This PR changes this behavior to make sure to use the custom `create_returnn_layer_dict`.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/module.py
+++ b/pytorch_to_returnn/torch/nn/modules/module.py
@@ -520,12 +520,20 @@ class Module:
     """
     Return True for modules which should be directly mapped to a RETURNN layer call. Otherwise, we assume that the
     module has custom forward code which will then be mapped to RETURNN layer calls on that level.
+
+    In order to determine, go through the class bases hierarchy and see which comes first, create_returnn_layer_dict or
+    forward.
     """
     if not cls.has_torch_forward():
       return True
-    for base in cls.__bases__:
-      if hasattr(base, "create_returnn_layer_dict") and cls.create_returnn_layer_dict == base.create_returnn_layer_dict:
+    base = cls
+    while base is not object:
+      if cls.create_returnn_layer_dict != base.create_returnn_layer_dict:
+        return True
+      elif cls.forward != base.forward:
         return False
+      assert len(base.__bases__) == 1, "Not implemented otherwise"
+      base = base.__bases__[0]
     return True
 
   def check_returnn_layer(self, layer: LayerBase):
--- a/pytorch_to_returnn/torch/nn/modules/module.py
+++ b/pytorch_to_returnn/torch/nn/modules/module.py
@@ -515,6 +515,19 @@ class Module:
       return True  # always assume that the user module has custom forward code, even if not cls.forward
     return cls.forward is not Module.forward
 
+  @classmethod
+  def direct_returnn_layer_call(cls) -> bool:
+    """
+    Return True for modules which should be directly mapped to a RETURNN layer call. Otherwise, we assume that the
+    module has custom forward code which will then be mapped to RETURNN layer calls on that level.
+    """
+    if not cls.has_torch_forward():
+      return True
+    for base in cls.__bases__:
+      if hasattr(base, "create_returnn_layer_dict") and cls.create_returnn_layer_dict == base.create_returnn_layer_dict:
+        return False
+    return True
+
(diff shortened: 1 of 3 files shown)


vieting:
I added `test_dct` as an example test case because that's what I was interested in. We can also simplify in general or reformat the code in `dct()` if you don't like it.

pytorch_to_returnn/torch/nn/modules/module.py
```diff
@@ -515,6 +515,19 @@ def has_torch_forward(cls) -> bool:
...
+    """
+    Return True for modules which should be directly mapped to a RETURNN layer call. Otherwise, we assume that the
+    module has custom forward code which will then be mapped to RETURNN layer calls on that level.
+    """
+    if not cls.has_torch_forward():
+      return True
+    for base in cls.__bases__:
+      if hasattr(base, "create_returnn_layer_dict") and cls.create_returnn_layer_dict == base.create_returnn_layer_dict:
+        return False
+    return True
```
albertz, line 529: This is a bit wrong and also too complicated. You can do it like in `has_torch_forward`, right?
```suggestion
    return cls.create_returnn_layer_dict is not Module.create_returnn_layer_dict
```

albertz:
There are three failing tests. Did you check them?

vieting:
> There are three failing tests. Did you check them?

Yes, they are the same as in #125.

pytorch_to_returnn/torch/nn/modules/module.py
vieting: This will work for standard cases, but not for e.g. what we have in `test_fp32_group_norm`. There, we have a module
```
class Fp32LayerNorm(torch.nn.LayerNorm):
  def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)

  def forward(self, input):
    output = F.layer_norm(
      input.float(),
      self.normalized_shape,
      self.weight.float() if self.weight is not None else None,
      self.bias.float() if self.bias is not None else None,
      self.eps
    )
    return output.type_as(input)
```
So in that case, `cls.create_returnn_layer_dict` is the same as `torch.nn.LayerNorm.create_returnn_layer_dict`, so no the same as `Module.create_returnn_layer_dict`. Still, we don't want to wrap it in a direct RETURNN layer call but rather wrap what is inside `forward`.

albertz: I don't understand. `LayerNorm.create_returnn_layer_dict` is the same as `Module.create_returnn_layer_dict`?

albertz: Ah sorry, it's the same as `_Norm. create_returnn_layer_dict`.

albertz: But then the logic is even more complex. You need to go through the class bases hierarchy and see which comes first, `forward` or `create_returnn_layer_dict`.

vieting: The idea is to always check whether the module implements `create_returnn_layer_dict` itself (not inherited). In this case, we should do a direct RETURNN layer call, otherwise not. This is the way to check it that I came up with, maybe there is a nicer one. I can also add comments to clarify. What do you think, isn't this a valid logic?

albertz: But having a module `Module` with whatever behavior (own `forward`, own `create_returnn_layer_dict`, or whatever), then this module:
```python
class DerivedModule(Module):
  pass
```
must have exactly the same behavior as `Module`. Otherwise it would be very unexpected.

vieting: I pushed a commit which changes the logic to what you suggested.

albertz: It probably still could be more generic but it's maybe fine for now.

albertz (approved):

PR #130, opened 2022-12-07
small additions

vieting:
Three small additions, should be trivial.


vieting:
The failing tests are the ones from #125. It's still unclear to me what causes them.

PR #132, opened 2022-12-07
add _FunctionalConvNd._get_output_shape_from_returnn

vieting:
`_FunctionalConvNd` is missing a `_get_output_shape_from_returnn()` method. It is very similar to the one of `_ConvNd`.


vieting:
The failing tests are the ones from https://github.com/rwth-i6/pytorch-to-returnn/issues/125.

albertz:
I wonder if this could share some code.

I also wonder whether this is still correct when you operate on other axes, as I mentioned in the other issue.

But anyway, fine for now.

PR #133, opened 2022-12-07
fix for tuples with non-hashable entries

vieting:
I had an issue with tuples that contain non-hashable entries, therefore I modified the check in `wrap()` to check if all entries are also hashable.


vieting:
The failing tests are the ones from #125.

albertz (approved):

Issue #134, opened 2022-12-07
Convolution over feature dim


vieting:
I have a case where the convolution is done over a dim, that RETURNN considers as feature dim. However, in pytorch a new dim is created before the convolution and this is now supposed to be the feature dim. Of course, RETURNN cannot directly know this. But I think in the case that a convolution is done over the feature dim and another static dim exists, we can argue that we should consider that other dim as feature dim and do the convolution. What do you think, @albertz?

A test to show the error:
```
def test_vgg():
  n_batch, n_time, n_feat = 3, 20, 5  # B, T, F

  def model_func(wrapped_import, inputs: torch.Tensor):
    if typing.TYPE_CHECKING or not wrapped_import:
      import torch
    else:
      torch = wrapped_import("torch")

    class VggBlock(torch.nn.Module):
      def __init__(self, n_in, n_out, kernel_size, pool_size=None, stride: typing.Union[int, typing.Tuple] = 1):
        super().__init__()
        self.conv = torch.nn.Conv2d(n_in, n_out, kernel_size, stride=stride)
        self.activation = torch.nn.SiLU()
        self.pooling = torch.nn.MaxPool2d(pool_size) if pool_size is not None else None

      def forward(self, x):
        # ignore padding here
        x = self.conv(x)
        x = self.activation(x)
        if self.pooling is not None:
          x = self.pooling(x)
        return x

    x = inputs.unsqueeze(1)  # (B, 1, T, F)
    # VGG block:
    vgg_blocks = torch.nn.Sequential(
      VggBlock(1, 32, (3, 3)),
    )
    x = vgg_blocks(x)
    x = x.transpose(2, 3).flatten(1, 2)  # (B, F, T)
    return x

  rnd = numpy.random.RandomState(42)
  x = rnd.normal(0., 1., (n_batch, n_time, n_feat)).astype("float32")
  converter = verify_torch_and_convert_to_returnn(
    model_func, inputs=x, returnn_dummy_input_shape=x.shape, validate_allclose_kwargs=dict(rtol=0, atol=5e-3),
    inputs_data_kwargs={"shape": (None, n_feat), "batch_dim_axis": 0, "time_dim_axis": 1, "feature_dim_axis": 2})
```

Error message:
```
...
  File ".../returnn/returnn/tf/layers/basic.py", line 5416, in ConvLayer.transform_input
[...]

PR #135, opened 2022-12-07
Convolution over feature dim

vieting:
As discussed in #134, it is currently not possible to do a convolution over the axis which RETURNN considers the feature dim and it would be helpful to set `in_dim`. However, then the basic `_get_output_shape_from_returnn` fails because there, the new feature dim is mapped to the old feature dim and as a result, the remaining dims are also mapped incorrectly. I add a suggestion in this PR.

### DIFF
--- a/pytorch_to_returnn/torch/nn/modules/conv.py
+++ b/pytorch_to_returnn/torch/nn/modules/conv.py
@@ -163,8 +163,8 @@ class _ConvNd(Module):
 
     if len(mapping_out_to_in) == in_data.batch_ndim - 1:
       # only one axis is missing, just take remaining axis
-      remaining_in = set(range(in_data.batch_ndim)).difference(set(mapping_out_to_in.keys()))
-      remaining_out = set(range(in_data.batch_ndim)).difference(set(mapping_out_to_in.values()))
+      remaining_in = set(range(in_data.batch_ndim)).difference(set(mapping_out_to_in.values()))
+      remaining_out = set(range(in_data.batch_ndim)).difference(set(mapping_out_to_in.keys()))
       assert len(remaining_in) == 1 and len(remaining_out) == 1
       mapping_out_to_in[remaining_out.pop()] = remaining_in.pop()
 
--- a/pytorch_to_returnn/torch/nn/modules/conv.py
+++ b/pytorch_to_returnn/torch/nn/modules/conv.py
@@ -132,8 +132,8 @@ class _ConvNd(Module):
                                      ) -> Tuple[Tuple[int, ...], Dict[int, int]]:
     """
     If the size of a dynamic axis changes (e.g. due to strides and/or padding), this is not covered in the base method
-    and we fix it here. Also, the basic returnn_axis_from_torch_axis fails if the RETURNN input feature dim is used as a
-    spatial dim for convolution. We try to cover this here and use the basic implementation as a fallback.
+    and we fix it here. Also, returnn_axis_from_torch_axis needs to be modified for cases where the RETURNN input
+    feature dim is used as a spatial dim for convolution.
     """
     assert len(inputs_flat) == 1
     torch_shape = list(inputs_flat[0].shape)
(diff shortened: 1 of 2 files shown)


vieting:
The failing tests are the ones from https://github.com/rwth-i6/pytorch-to-returnn/issues/125.

pytorch_to_returnn/torch/nn/modules/conv.py
```diff
@@ -78,14 +78,24 @@ def create_returnn_layer_dict(self, input: Tensor) -> Dict[str, Any]:
     assert all(p == 0 for p in self.padding)  # not implemented otherwise
     assert all(p == 0 for p in self.output_padding)  # not implemented otherwise
     assert self.padding_mode == "zeros"  # not implemented otherwise
+
+    from pytorch_to_returnn.naming import Naming
+    naming = Naming.get_instance()
+    input_tensor = naming.tensors[input]
+    in_dim = input_tensor.returnn_data.dim_tags[input_tensor.returnn_axis_from_torch_axis[1]]
+    in_spatial_dims = [
+      input_tensor.returnn_data.dim_tags[input_tensor.returnn_axis_from_torch_axis[dim + len(input.shape)]]
```
albertz, line 87: Why don't you use `_get_input_axis_to_returnn` for that?

pytorch_to_returnn/torch/nn/modules/conv.py
```diff
@@ -121,18 +131,53 @@ def import_params_torch_to_returnn(self, *, layer: LayerBase, torch_module: _Con
...
-      inputs_flat=inputs_flat, layer=layer)
     assert len(inputs_flat) == 1
     torch_shape = list(inputs_flat[0].shape)
     torch_shape[1] = self.out_channels
     for idx in range(self.nd):
       torch_ax = idx + 2
       torch_shape[torch_ax] = (torch_shape[torch_ax] + 2 * self.padding[idx] - self.dilation[idx] * (
         self.kernel_size[idx] - 1) - 1) // self.stride[idx] + 1
+
+    from pytorch_to_returnn.naming import Naming
```
albertz, line 146: I don't understand why you do it so complicated. This code here should be quite short, straightforward, and not using any heuristics. Your code here is full of heuristics, checking whether you can map all axes, etc. You don't need that. We know exactly how it must map.

vieting: > We know exactly how it must map.

All dims that did not change are trivial, the channel dim can be done as I do it here. How would you do it for the spatial dims? Just assume that the order of spatial dims is the same as in `in_spatial_dims`?

pytorch_to_returnn/torch/nn/modules/conv.py
vieting: Because that would return an axis description like `"T"` or `"F"`. This would be mapped to a dim tag in the `ConvLayer` construction. However, in case we do convolution over the feature dim, `"F"` would be mapped to the `in_dim`, so the new feature dim and not the old one which does not work.

pytorch_to_returnn/torch/nn/modules/conv.py
vieting: I added an update which does the mapping as I described above.

albertz: > Just assume that the order of spatial dims is the same as in `in_spatial_dims`?

What order? Of the RETURNN output?

We don't need to guess anything here. We know everything exactly.

albertz: Except BCHW vs BHWC, but you can just check where out_dim is.

vieting: Ok, then that is exactly what I do now, right?

vieting:
If we do this change, I have a follow up issue. The test cases work, but if I write the conversion result to a file via `converter.get_returnn_config_serialized()`, the dim tags will show up at the beginning and are used in the network as well. Here is an example:

<details>

```
import numpy
from returnn.tf.util.data import Dim, batch_dim, single_step_dim, SpatialDim, FeatureDim

use_tensorflow = True
behavior_version = 12

feature_data_dim = FeatureDim('feature:data', 11)
time_data_dim = SpatialDim('time:data')
spatial1_data_dim = SpatialDim('spatial1:data')
Linear_feature_dense_dim = FeatureDim('Linear:feature-dense', 7)

extern_data = {
  'data': {
    'dim_tags': [
      batch_dim,
      feature_data_dim,
      time_data_dim,
      spatial1_data_dim
    ],
    'dtype': 'float32',
    'time_dim_axis': 2,
    'feature_dim_axis': 1
  }
}

network = {
  'Transpose': {'class': 'copy', 'from': 'data'},
  'Linear': {
    'class': 'linear',
    'from': 'Transpose',
    'n_out': 7,
    'with_bias': True,
    'activation': None
  },
  'Transpose_1': {'class': 'copy', 'from': 'Linear'},
  'Conv2d': {
    'class': 'conv',
    'from': 'Transpose_1',
    'activation': None,
    'with_bias': True,
    'n_out': 13,
    'filter_size': (3, 5),
    'padding': 'valid',
    'in_spatial_dims': [
      time_data_dim,
      spatial1_data_dim
    ],
    'in_dim': Linear_feature_dense_dim,
    'strides': (2, 2)
  },
  'output': {'class': 'copy', 'from': 'Conv2d'}
}
```

</details>

However, `Linear_feature_dense_dim` in the example is not identical with the dim tag in the input. We would have to set the output dim tags also I guess, right? Are there already solutions for this in returnn_common or elsewhere?

albertz:
You need to change all `n_out` to `out_dim`. This is for `LinearLayer` but also other layers like `ConvLayer`.

You also need to specify `out_spatial_dims` for `ConvLayer`.

vieting:
I need to create them in `create_returnn_layer_dict` and there is no way to infer them, right?

And the printed config will possibly have LOTS of dim tags at the beginning. I guess that's similar in returnn common, right?

